PR #49, opened 2023-06-02
Mlflow server

matbun:
Switch to HTTP communication with MLFlow server. Easily spawn/kill MLFLow server on localhost, by defininf new steps in the DT workflows


matbun:
@MrGweep this is a working example of how to use MLFLow server with HTTP communication. See if/how you want to merge

MrGweep (changes requested):
Tried it out and works fine. One test failed cause of a type, so I suggested it in the comments.

MrGweep (approved):

MrGweep (approved):

PR #55, opened 2023-06-29
Fix pytorch env path

matbun:


matbun:
@MrGweep it seems there is a small problem with CWL

MrGweep:
> @MrGweep it seems there is a small problem with CWL

Just had a quick look. Apparently it doesn't find the .cwl files and calls it with `cwltool None` for some reason idk yet. Does this block you atm @matbun?

matbun:
> > @MrGweep it seems there is a small problem with CWL
> 
> Just had a quick look. Apparently it doesn't find the .cwl files and calls it with `cwltool None` for some reason idk yet. Does this block you atm @matbun?

No, I was just doing some housekeeping. But this error may suggest that cwl workflow is not fully working, for some reason

PR #90, opened 2023-11-01
Migrate to containerized workflows for MNIST (torch)

MrGweep:
Workflow steps are currently Python environments. To integrate with the infrastructure, they had to be converted into containers.

Three ways of submitting a job to SLURM using containers have been implemented:
1. Direct build on HPC via Apptainer (tested)
2. Converting pre-built Docker to Apptainer on HPC (tested)
3. submit to Kubernetes and offload to HPC via Interlink (currently under testing)

Please check the Readme for a more detailed explanation. A startscript has been adapted and included in `use-cases/mnist/torch/startscript.sh`

A GitHub action has been created to build the Docker container and push it the repository automatically. However, this is currently not used due to the large size of the containers exceeding the limit of 14GB memory for a GitHub action runner. To solve this future work could set up our own GitHub Action runner.


.github/workflows/build-docker.yml
```diff
@@ -0,0 +1,50 @@
+#
+name: Create and publish a Docker image
+
+# Configures this workflow to run every time a change is pushed to the branch called `release`.
+on:
+  push:
+    branches:
+          -  containers
```
matbun, line 8: The branch should be main, to prevent pushing images under development to the containers registry. Maybe we could have a workflow for main and dev branches ("-dev" in the image name should be self-explanatory)

.github/workflows/build-docker.yml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+      # This step uses the `docker/build-push-action` action to build the image, based on your repository's `Dockerfile`. If the build succeeds, it pushes the image to GitHub Packages.
+      # It uses the `context` parameter to define the build's context as the set of files located in the specified path. For more information, see "[Usage](https://github.com/docker/build-push-action#usage)" in the README of the `docker/build-push-action` repository.
+      # It uses the `tags` and `labels` parameters to tag and label the image with the output from the "meta" step.
+      - name: Build and push Docker image
+        uses: docker/build-push-action@f2a1d5e99d037542a71f64918e516c093c6f3fc4
+        with:
+          context: .
+          file: containers/Dockerfile
```
matbun, line 47: see below

containers/Dockerfile
matbun, line 1: For the sake of having a simple repo structure, would it make more sense to move this under "env-files" folder? Or the containers folder is expected to grow?

containers/apptainer/apptainer_build_env.sh
```diff
@@ -0,0 +1,9 @@
+nname='torch_env'
+
+# create env inside container
+# python3 -m venv $nname --system-site-packages
+# source ${nname}/bin/activate
+
+# install wheels -- from this point on, feel free to add anything
+#pip3 install -r ./env-files/torch/pytorch-env-gpu-container.txt
```
matbun, line 8: can be safely removed

docs/docs/img/containers.png
matbun, line 1: need to agree on a common name for startscript files. ATM we were using just "startscript"

env-files/torch/pytorch-env-gpu-container.txt
matbun, line 1: This can be safely removed

pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "itwinai"
 version = "0.1"
 description = "AI and ML workflows module for interTwin"
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = ">=3.8"
```
matbun, line 10: is python 3.8 needed?

pyproject.toml
```diff
@@ -32,8 +32,8 @@ dependencies = [
     "pyyaml>=6.0.1",
     "omegaconf>=2.3.0",
     "submitit>=1.4.6",
-    "typing-extensions==4.5.0",
-    "typing_extensions==4.5.0",
+    "typing-extensions>=4.5.0",
```
matbun, line 35: These '==' were there because of some bug introduced in later versions. But we can try to remove/comment these lines, as the bug may have been solved by now

pyproject.toml
```diff
@@ -32,8 +32,8 @@ dependencies = [
...
+    "typing_extensions>=4.5.0",
```
matbun, line 36: These '==' were there because of some bug introduced in later versions. But we can try to remove/comment these lines, as the bug may have been solved by now

tests/use-cases/conftest.py
```diff
@@ -7,7 +7,7 @@
 
 FNAMES = [
     'pipeline.yaml',
-    'startscript',
+    'startscript.sh',
```
matbun, line 10: need to agree on a common name for startscript files. ATM we were using just "startscript". better to have only one

use-cases/mnist/torch/startscript.sh
matbun, line 1: need to agree on a common name for startscript files. ATM we were using just "startscript"

matbun (changes requested):
@r-sarma how do you think we should call startscript files? If I well remember, you care calling them "startscript". This could be a nice opportunity to define a "standard" among ourselves.

Some tests are failing, partially due to code formatting and to inconsistent use case folder structure (startscript name)

r-sarma:
> @r-sarma how do you think we should call startscript files? If I well remember, you care calling them "startscript". This could be a nice opportunity to define a "standard" among ourselves.
> 
> Some tests are failing, partially due to code formatting and to inconsistent use case folder structure (startscript name)

'startscript' is a standard name, but can be also different. Maybe LaunchSLURM, or LaunchJob? I am open to changing it.

pyproject.toml
MrGweep: 3.9 is not available for the NGC container

pyproject.toml
MrGweep: same reason: needed for the NGC container

containers/apptainer/apptainer_build_env.sh
MrGweep: removed

.github/workflows/build-docker.yml
MrGweep: removed GH actions, since it is not used currently

.github/workflows/build-docker.yml
MrGweep: removed GH actions, since it is not used currently

README.md
matbun, line 1: Suggestion: you could create a docs page under `docs/`, so that it gets automatically published next time we merge into main branch

containers/Dockerfile
matbun, line 1: as discussed before, consider moving the `containers/` folder under `env-files/` for a better organization of the repo

env-files/torch/pytorch-env-gpu-container.txt
matbun, line 1: this should be moved to `pyproject.toml`

tests/use-cases/conftest.py
matbun: if this changes, you need to propagate the change to all the startscripts in the repo accordingly

matbun (changes requested):
startscript names need to be consistent (all "startscript" or all "startscript.sh") to solve failing tests

tests/use-cases/conftest.py
r-sarma: If it is currently startscript elsewhere, let us stick to that without '.sh'

use-cases/mnist/torch/startscript.sh
```diff
@@ -0,0 +1,102 @@
...
+#         $CONTAINERPATH \
+#         $EXEC
+# fi
+
+#Choose SLURM execution or bash script execution
+if [ "$SLURM_EXECUTION" = true ]; then
+    srun --cpu-bind=none bash -c "apptainer run --nv \
+        $CONTAINERPATH"
+
+else
```
r-sarma, line 94: Why would you not use SLURM here?

matbun:
Solved by https://github.com/interTwin-eu/itwinai/pull/197

PR #98, opened 2023-11-10
3dgan integration

matbun:
Update 3DGAN training


use-cases/3dgan/Dockerfile.vega
MrGweep, line 1: Maybe a more descriptive name like Dockerfile-inference-cpu is better

MrGweep (changes requested):

MrGweep (approved):

PR #130, opened 2024-04-29
Cyclone tf dist

MarioRuettgers:
Integration of cyclone use case in compliance with tf distributed strategy


matbun:
I will take care of fixing the conflicts

PR #136, opened 2024-05-03
Python venv

matbun:
Move away from micromamba and provide a generic way to create virtual environments, which is independent from JSC and micromamba. Also, update GH actions and integration tests. Also, add automatic publish to PyPI when a new release is created.


matbun:
If there are no comments I will merge this by tomorrow EOB

PR #137, opened 2024-05-07
Distributed strategy launcher

matbun:
Move away from micromamba and provide a generic way to create virtual environments, which is independent from JSC and micromamba. Also, update GH actions and integration tests. Also, add automatic publish to PyPI when a new release is created.

This PR is similar to #136 , including in addition new/updated TF tutorials (e.g., JUBE).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ name = "itwinai"
 version = "0.0.1"
 description = "AI and ML workflows module for scientific digital twins."
 readme = "README.md"
-requires-python = ">=3.11"
+requires-python = ">=3.10"
 license = { file = "LICENSE" }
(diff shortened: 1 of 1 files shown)


env-files/tensorflow/createEnvJSCTF.sh
```diff
@@ -69,28 +62,10 @@ if [ "$cont1" = true ] ; then
   fi
 fi
 
-# install TF
```
r-sarma, line 72: Why is this removed?

r-sarma: Ok I see that you moved the rest of it to the other file. But why not also keep the Horovod setup?

r-sarma (approved):

PR #167, opened 2024-06-24
Update generic_tf.sh

r-sarma:
# Summary

TF version fixing needs to be in the format 2.16.*

**Related issue :**


matbun (approved):
LGTM. I would suggest merging this also into main

PR #193, opened 2024-07-25
fix gan bug and update docs

MutegekiHenry:
# Summary

## This PR fixes the following:

- Removes the `log_all_workers` parameter in the GAN tutorial as it was removed from the TorchTrainer class.
- Corrects the train file name in the GAN readME to `train.py`
- Corrects the torch-tutorial0-basics ReadME on the instructions to run horovod using slurm. 

---

**Related issue :**


matbun (approved):
LGTM

PR #195, opened 2024-08-15
Update general_jobsys.xml

r-sarma:
Fix to JUBE benchmarking code

---

**Related issue :**


matbun (commented):
It seems that there are no changes to be made into the main. Also, some unit tests are failing

PR #207, opened 2024-09-13
Stylistic changes

jarlsondre:
# Summary
Small stylistic changes to make code and tutorials more readable.

---

**Related issue :**

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,3 +1,0 @@
-[settings]
-profile = black
-known_first_party = itwinai
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,10 @@
+[settings]
+profile = black
+# multi_line_output = 3
+# include_trailing_comma = true
+# force_grid_wrap = 0
+# use_parentheses = true
+# line_length = 88
+# force_sort_within_sections = true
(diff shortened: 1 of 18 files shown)


matbun (commented):
Thank you for the fixes!
Since you brought up the problem of imports orderings, why not complying with some "standard"? You could run [isort](https://pycqa.github.io/isort/) on the whole package and *ideally* see if it can be activated in our [superlinter](https://github.com/super-linter/super-linter) CI/CD linter on GH. Alternativelly you can look for isort GH actions.

jarlsondre:
@matbun I have now added support for isort and fixed the files that it complained about initially. It seems that when running `isort <filename>` from the command line without specifying anything else that it will not use the same style as the super-linter does. I could not seem to figure out which "profile" the super-linter uses for isort, but when adding "black" as the profile it seemed to work for the initial files that were tested. Therefore I added "black" as the profile in the `pyproject.toml` file. There are other profiles to choose from here, though: https://pycqa.github.io/isort/docs/configuration/profiles.html

matbun (approved):
LGTM

.isort.cfg
matbun, line 1: Could you see if this could be embedded into the `pyproject.toml` please? It would reduce cluttering of the repo's root

jarlsondre: Fixed this now :)

matbun (approved):

PR #222, opened 2024-10-09
Distributed print decorator

matbun:
Update the `monior_exec` decorator to suppress prints on workers with global rank != 0 only when *explicitly* asked to do so.

Actually, the logic meant to suppress the print on workers  with global rank != 0 has been moved to another decorator, which makes its usage straightforward. Example:

```python
from itwinai.distributed import suppress_workers_print

class SilentComponent(BaseComponent):
    """Fake component class to test the decorator meant to suppress print()."""

    def __init__(self, max_items: int, name: str = 'SilentComponent') -> None:
        super().__init__(name)
        self.save_parameters(max_items=max_items, name=name)
        self.max_items = max_items

    @suppress_workers_print
    def execute(self, foo, bar=123):
        print("Executing SilentComponent")
        return "silent component result"
```

The two decorators also can be used together. Example:

````python
class FakeComponent(BaseComponent):
    """Fake component class to test the use of two decorators at the same time."""

    def __init__(self, max_items: int, name: str = 'FakeComponent') -> None:
        super().__init__(name)
        self.save_parameters(max_items=max_items, name=name)
        self.max_items = max_items

    @suppress_workers_print
    @monitor_exec
    def execute(self, foo, bar):
        print("Executing FakeComponent")
        return "fake component result"

```

### DIFF
--- a/src/itwinai/components.py
+++ b/src/itwinai/components.py
@@ -94,7 +94,7 @@ from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 from .distributed import (
     builtin_print,
     detect_distributed_environment,
-    distributed_patch_print
+    distributed_patch_print,
 )
 from .serialization import ModelLoader, Serializable
 from .type import MLArtifact, MLDataset, MLModel
--- a/src/itwinai/components.py
+++ b/src/itwinai/components.py
@@ -116,46 +116,6 @@ def monitor_exec(method: Callable) -> Callable:
         return result
     return wrapper
 
-# def monitor_exec(print_on_main_worker_only: bool = False) -> Callable:
-#     """Decorator for execute method of a component class.
-#     Computes execution time and gives some information about
-#     the execution of the component.
-
-#     Args:
-#         print_on_main_worker_only (bool): whether to ignore ``print()`` calls on
-#             workers which are not the main worker in a distributed context,
-#             such as distributed ML. If set to ``True``, to force printing on
-#             all workers you will need to use ``print(..., force=True)``.
-#     """
-
-#     def method_decorator(method: Callable) -> Callable:
-#         @functools.wraps(method)
-#         def wrapper(self: BaseComponent, *args, **kwargs) -> Any:
-#             if print_on_main_worker_only:
-#                 # Disable print in workers different from the main one,
-#                 # when in distributed environments.
-#                 dist_grank = detect_distributed_environment().global_rank
-#                 patched_print = distributed_patch_print(is_main=dist_grank == 0)
-#                 __builtin__.print = patched_print
-
-#             msg = f"Starting execution of '{self.name}'..."
-#             self._printout(msg)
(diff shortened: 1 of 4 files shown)


src/itwinai/components.py
```diff
@@ -88,47 +88,58 @@
...
+            try:
+                result = method(self, *args, **kwargs)
+            finally:
+                self.cleanup()
+            self.exec_t = time.time() - start_t
+            msg = f"'{self.name}' executed in {self.exec_t:.3f}s"
+            self._printout(msg)
+
+            # Reset print() to __builtins__ one
+            __builtin__.print = builtin_print
```
jarlsondre, line 136: I am kind of new to this syntax here, but wouldn't this either 
1) Reset what you did in the if statement or
2) Repeat what you did in the if statement (thus giving no value)?

Not sure what applies in terms of scope etc. here :)

matbun: it will reset it, avoiding to tamper with prints in subsequent code. We only want to ignore prints on workers for the component's logic

jarlsondre: I don't think I am entirely following. Wouldn't that negate the stuff you did on line 119 and below?

matbun: yes

jarlsondre: Could you not just use something similar to "strategy.is_main_worker" inside of this function to do this, instead of overwriting the built-in print function? Since it is only for this scope anyway, why overwrite the global function?

matbun: what you suggest would be possible only when strategy is available, which is not always true for all components unfortunately. You may also mean "why not checking if strategy.is_main_worker every time before printing": this is possible but tedious and error prone. The new version of the code should solve both problems providing two decorators, one to monitor the execution of a component, the other to suppress the prints. It is now much more straightforward

jarlsondre: I see now that I misunderstood what you were trying to do. I didn't consider the fact that you are calling the component before resetting, so I thought you were suppressing only the 2-3 prints where you are saying that training started. Now I see that you are also suppressing all the calls in the component as well, and so now I fully agree with what you are doing.

src/itwinai/components.py
```diff
@@ -84,51 +84,37 @@
...
-    Args:
-        func (Callable): class method.
     """
-    @functools.wraps(method)
-    def monitored_method(self: BaseComponent, *args, **kwargs) -> Any:
-        # Disable print in workers different from the main one,
-        # when in distributed environments.
-        dist_grank = detect_distributed_environment().global_rank
-        distributed_patch_print(is_main=dist_grank == 0)
+    Decorator for ``BaseComponent``'s methods.
```
jarlsondre, line 99: Remove the newline after '"""' to follow docstring convention

jarlsondre:
Everything looks good to me :)

Issue #221, opened 2024-09-27
Updated scalability report (more comprehensive and easier to use)


jarlsondre:
# Summary
The scalability report that you get from `itwinai scalability_report` shows the relative speedup between running a job on a single node vs. on multiple nodes. However, this is not very comprehensive and has some issues with the user experience. 

## Metrics that should be included
- Throughput
  - FLOPS, samples/sec, or just total runtime (as this will be inversely proportional)
- Communication Overhead
  - How much time is spent performing communication vs. actual computation
- GPU Utilization
  - How much of the GPUs is utilized e.g. on average (as a score from 0 to 100%)   

## Other improvements
- Allowing to pass a folder without needing a RegEx pattern
  - Could still allow the use of RegEx, but if you know that a folder contains only the log files then you should be able to just specify the folder without having to think about patterns. If you choose to pass a RegEx then it will be used inside of the given folder e.g. 
- Perhaps a user interface such as streamlit or tensorboard? 
  - The PyTorch profiler is supposed to have an integration with tensorboard, so could be worth looking into.

jarlsondre:
# Suggested solution
Based on the literature around Horovod, DDP and DeepSpeed (as well as a couple of others, like OneFlow), it seems that most papers focus on throughput, measured as samples/sec or FLOPS, and that a select few (e.g. Horovod) seem to put some emphasis into GPU Utilization and time spent on communication vs computation. Since wall-clock time should be inversely proportional to throughput and is much easier to measure, I suggest we use this. 

**Note**: Any data displayed in the following plots will be completely fabricated by me, so don't read into the numbers. 

## Throughput

We measure the scalability of throughput in two ways: 
- Relative to time spent on a single node (or GPU), e.g. 
![image](https://github.com/user-attachments/assets/3b72a0c1-48c3-4908-99e2-57f939db6026)
- Absolute time, e.g. 
![image](https://github.com/user-attachments/assets/435a570a-fc39-4020-9af7-a7922bf742ed)

## Communication vs. Computation

We measure communication vs. computation as a score from 0 to 1, where 0 means all the time was spent on communication and 1 means that all the time was spent on computation. An example can be seen here: 
![image](https://github.com/user-attachments/assets/fc3e41d1-2cc4-4df0-8bb2-4d7f22128edd)
**Note**: The numbers 4, 8 and 16 refer to the number of GPUs in this plot. This is just a draft :) 

## GPU Utilization
TODO

matbun:
[...]

PR #227, opened 2024-10-16
Scalability report update - Communication Plot

jarlsondre:
# Summary

This PR is focused on improving the scalability report. In particular, this PR is centered around the communication plot, as discussed in #221, and provides two main features: 

1) A decorator that can be put on top of a `TorchTrainer` instance to do profiling. Any subclass must add `self.profiler.step()` in their `set_epoch()` function for this to work as intended. 

2) A CLI command that generates a communication plot in the folder that you are currently in. 

---

**Related issue :**
This is related to issue #221


src/itwinai/torch/communication_plot.py
matbun, line 1: Maybe move this file under itwinai.torch.profiling (new) package?

src/itwinai/torch/profiler.py
matbun, line 1: Maybe move this file under itwinai.torch.profiling (new) package?

matbun:
https://github.com/interTwin-eu/itwinai/pull/231 should be a cleanup of this

PR #236, opened 2024-10-24
Itwinai jlab Docker image

matbun:
Add Dockerfile and supporting files to generate jupyterLab singleuser container to be offloaded through interLink


README.md
```diff
@@ -346,14 +346,15 @@ For example, in `ghcr.io/intertwin-eu/itwinai:0.2.2-torch2.6-jammy`:
 The `TAG` follows the convention:
 
 ```text
-X.Y.Z-[torch|tf]x.y-distro
+[jlab-]X.Y.Z-(torch|tf)x.y-distro
```
jarlsondre, line 349: In this case, `(torch|tf)` is an "XOR", right? You can pick one or the other, but not both? Not super important but could consider mentioning it?

env-files/tensorflow/generic_tf.sh
```diff
@@ -88,9 +88,9 @@ pip3 install --no-cache-dir  tf_keras==2.16.*
 
 # Install Pov4ML
 if [[ "$OSTYPE" =~ ^darwin ]] ; then
-  pip install "prov4ml[apple]@git+https://github.com/matbun/ProvML@main" || exit 1
+  pip install "prov4ml[apple,nvidia]@git+https://github.com/matbun/ProvML@new-main" || exit 1
 else
-  pip install "prov4ml[linux]@git+https://github.com/matbun/ProvML@main" || exit 1
+  pip install "prov4ml[linux,nvidia]@git+https://github.com/matbun/ProvML@new-main" || exit 1
```
jarlsondre, line 93: I might be mistaken here, but from what I can tell, the `new-main` branch doesn't have an extra called `linux` anymore. The `linux` in the `main` branch seems to contain what `amd` contains in the `new-main` branch.

env-files/tensorflow/generic_tf.sh
```diff
@@ -88,9 +88,9 @@ pip3 install --no-cache-dir  tf_keras==2.16.*
 
 # Install Pov4ML
 if [[ "$OSTYPE" =~ ^darwin ]] ; then
-  pip install "prov4ml[apple]@git+https://github.com/matbun/ProvML@main" || exit 1
+  pip install "prov4ml[apple,nvidia]@git+https://github.com/matbun/ProvML@new-main" || exit 1
```
jarlsondre, line 91: This might be fine, but I do wonder how it is possible for `apple` and `nvidia` to be compatible extras here.

env-files/torch/Dockerfile
```diff
@@ -1,31 +1,70 @@
...
+    HOROVOD_GPU_ALLREDUCE=NCCL \
+    HOROVOD_NCCL_LINK=SHARED \
+    # DeepSpeed
+    DS_BUILD_UTILS=1 \
+    DS_BUILD_AIO=1 \
+    DS_BUILD_FUSED_ADAM=1 \
+    DS_BUILD_FUSED_LAMB=1 \
+    DS_BUILD_TRANSFORMER=1 \
+    DS_BUILD_STOCHASTIC_TRANSFORMER=1 \
+    DS_BUILD_TRANSFORMER_INFERENCE=1
```
jarlsondre, line 44: Is there any particular reason for not including `DS_BUILD_CCL_COMM` here? It is included in `generic_torch.sh`.

env-files/torch/Dockerfile
```diff
@@ -1,31 +1,70 @@
...
+
+# DeepSpeed, Horovod and other deps
+ENV HOROVOD_WITH_PYTORCH=1 \
+    HOROVOD_WITHOUT_TENSORFLOW=1 \
+    HOROVOD_WITHOUT_MXNET=1 \
+    CMAKE_CXX_STANDARD=17 \
+    HOROVOD_MPI_THREADS_DISABLE=1 \
+    HOROVOD_CPU_OPERATIONS=MPI \
+    HOROVOD_GPU_ALLREDUCE=NCCL \
+    HOROVOD_NCCL_LINK=SHARED \
```
jarlsondre, line 36: I notice that these are not the exact same envs that are in `generic_torch.sh`. Is there any particular reason for the difference?

env-files/torch/generic_torch.sh
```diff
@@ -179,9 +157,9 @@ fi
 
 # Install Pov4ML
 if [[ "$OSTYPE" =~ ^darwin ]] ; then
-  pip install --no-cache-dir "prov4ml[apple]@git+https://github.com/matbun/ProvML" || exit 1
+  pip install --no-cache-dir "prov4ml[apple,nvidia]@git+https://github.com/matbun/ProvML@new-main" || exit 1
 else
-  pip install --no-cache-dir "prov4ml[linux]@git+https://github.com/matbun/ProvML" || exit 1
+  pip install --no-cache-dir "prov4ml[linux,nvidia]@git+https://github.com/matbun/ProvML@new-main" || exit 1
```
jarlsondre, line 162: same as above with the linux/nvidia flags

env-files/torch/jupyter/Dockerfile
```diff
@@ -0,0 +1,256 @@
...
+    DS_BUILD_FUSED_ADAM=1 \
+    DS_BUILD_FUSED_LAMB=1 \
+    DS_BUILD_TRANSFORMER=1 \
+    DS_BUILD_STOCHASTIC_TRANSFORMER=1 \
+    DS_BUILD_TRANSFORMER_INFERENCE=1
+RUN pip install --no-cache-dir \
+    deepspeed==0.15.* \
+    git+https://github.com/horovod/horovod.git@3a31d93 \
+    "prov4ml[linux,nvidia]@git+https://github.com/matbun/ProvML@e99bddba047a1a04ca0387e87cad1ed96d5ab2c5" \
+    ray ray[tune]
```
jarlsondre, line 237: isn't it sufficient here to just do `ray[tune]`?

env-files/torch/jupyter/configure.py
```diff
@@ -0,0 +1,138 @@
...
+# HOME = '/home/jovyan'
+# HOME = '/ceph/hpc/home/ciangottinid'
+
+
+def write_jupyterlab_config():
+    HOME = os.getenv('HOME', '/ceph/hpc/home/ciangottinid')
+
+    file_path = HOME + '/.jupyter/jupyter_notebook_config.json'
+    if not os.path.isfile(file_path):
+        os.makedirs(HOME + '/.jupyter/', exist_ok=True)
```
jarlsondre, line 18: This should work, but I would use pathlib here for best practice

env-files/torch/jupyter/configure.py
```diff
@@ -0,0 +1,138 @@
...
+        'request_retries': 3,
+        'protocol_stat_retries': 6
+    }
+    client_config = dict((k, v) for k, v in client_config.items() if v)
+    rucio_config['client'] = client_config
+
+    if not os.path.isfile('/opt/rucio/etc/rucio.cfg'):
+        os.makedirs('/opt/rucio/etc/', exist_ok=True)
+
+    with open('/opt/rucio/etc/rucio.cfg', 'w') as f:
```
jarlsondre, line 132: Same as above, pathlib

tests/torch/runall.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+# Python virtual environment
+PYTHON_VENV="../../.venv-pytorch"
+
+# Clear SLURM logs (*.out and *.err files)
+rm -rf logs_slurm
+mkdir logs_slurm
```
jarlsondre, line 8: would `mkdir -p logs_slurm` not work here? If any of the files already exist then they will be overwritten anyway, and the rm will throw an error (though not terminate the program so that's okay I guess) if this folder doesn't already exist.

tests/torch/runall.sh
```diff
@@ -0,0 +1,49 @@
...
+export NO_COLOR=1
+
+export DIST_MODE="ddp"
+export RUN_NAME="ddp-itwinai"
+export COMMAND='pytest -v -m torch_dist'
+sbatch  \
+    --job-name="$RUN_NAME-n$N" \
+    --output="logs_slurm/job-$RUN_NAME-n$N.out" \
+    --error="logs_slurm/job-$RUN_NAME-n$N.err" \
+    slurm.vega.sh
```
jarlsondre, line 31: This might be neater if separated into a small function

src/itwinai/torch/distributed.py
```diff
@@ -613,24 +564,28 @@ def init(self) -> None:
         import deepspeed
 
         self.deepspeed = deepspeed
-        if not distributed_resources_available():
-            raise RuntimeError("Trying to run distributed on insufficient resources.")
+        force_dist_env = int(os.environ.get('ITWINAI_FORCE_DIST', '0'))
+        if not distributed_resources_available() and not force_dist_env:
```
annaelisalappe, line 568: Maybe to make this a little more neat we could move the `force_dist_env = int(os.environ.get('ITWINAI_FORCE_DIST', '0'))` to the `distributed_resources_available()` function so as to avoid duplication, and instead call it something like `should_run_distributed()`. Just a design preference, both would work obviously

tests/torch/distribtued_decorator.py
annaelisalappe, line 1: file name :)

tests/torch/runall.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Python virtual environment
+PYTHON_VENV="../../.venv-pytorch"
+
+# Clear SLURM logs (*.out and *.err files)
+rm -rf logs_slurm
+mkdir logs_slurm
+rm -rf logs_torchrun logs_mpirun logs_srun
+
+export MNIST_PATH="/ceph/hpc/data/st2301-itwin-users/mbunino/mnist" #"/p/project1/intertwin/smalldata/mnist"
```
annaelisalappe, line 11: I guess these file paths are meant to be specific to mbunino because of the way the Vega file structure is set up, right?

tests/torch/test_distribtued.py
annaelisalappe, line 1: file name here also :)

env-files/tensorflow/generic_tf.sh
matbun: Good point. Can you install `nvitop` on apple devices? If not, I will remove it

tests/torch/test_distribtued_training.py
annaelisalappe, line 1: file name also here

env-files/torch/Dockerfile
matbun: the differences are irrelevant in this case

env-files/torch/Dockerfile
```diff
@@ -1,31 +1,70 @@
...
-USER jovyan:jovyan
-
-LABEL org.opencontainers.image.source=https://github.com/interTwin-eu/itwinai
-LABEL org.opencontainers.image.description="Base itwinai image with torch dependencies and CUDA drivers"
-LABEL org.opencontainers.image.licenses=MIT
-LABEL maintainer="Matteo Bunino - matteo.bunino@cern.ch"
+COPY pyproject.toml pyproject.toml
+COPY src src
+# Torch: reuse the global torch in the container
+RUN CONTAINER_TORCH_VERSION="$(python -c 'import torch;print(torch.__version__)')" \
```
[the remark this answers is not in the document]
matbun: actually the NGC container already has torch in the container, so I am reusing it

env-files/torch/jupyter/Dockerfile
```diff
@@ -0,0 +1,256 @@
...
+RUN git clone https://github.com/NVIDIA/apex && cd apex \
+    && pip install -v --disable-pip-version-check --no-cache-dir --no-build-isolation --config-settings "--build-option=--cpp_ext" --config-settings "--build-option=--cuda_ext" ./ \
+    && rm -rf /tmp/apex
+# Transformer engine: https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/index.html
+# (needed for DeepSpeed *TRANSFORMER* optinal build options)
+# Note: it will take about half an hour to build
+RUN pip install --no-cache-dir transformer_engine[pytorch]
+
+# DeepSpeed, Horovod and other deps
+ENV HOROVOD_WITH_PYTORCH=1 \
```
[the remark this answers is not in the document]
matbun: only LDSHARED and HOROVOD_NCCL_HOME are not set in the Dockerfile, which is ininfluent in this case

env-files/torch/jupyter/configure.py
matbun: that's just a file I copied from another repo actually, which I am not maintaining

tests/torch/runall.sh
matbun: this will not delete the directory. Anyways, this script is not so important

tests/torch/test_torch_trainer.py
```diff
@@ -0,0 +1,123 @@
...
+        config=training_config,
+        epochs=2,
+        strategy='horovod',
+        checkpoint_every=1
+    )
+    # Patch strategy with already initialized one to avoid re-initialization
+    trainer.strategy = horovod_strategy
+
+    train_set, val_set = mnist_datasets
+    trainer.execute(train_set, val_set)
```
annaelisalappe, line 123: You could consider parametrising these tests, like this: 
```
@pytest.mark.hpc
@pytest.mark.parametrize(
    'strategy_name,strategy_fixture',
    [
        ('ddp', 'ddp_strategy'),
        ('deepspeed', 'deepspeed_strategy'),
        ('horovod', 'horovod_strategy')
    ]
)
def test_distributed_trainer_mnist(mnist_datasets, request, strategy_name, strategy_fixture):
    """Test TorchTrainer on MNIST with different distributed strategies."""
    training_config = dict(
        optimizer='sgd',
        loss='nllloss'
    )
    trainer = TorchTrainer(
        model=Net(),
        config=training_config,
        epochs=2,
        strategy=strategy_name,
        checkpoint_every=1
    )
   
    strategy_instance = request.getfixturevalue(strategy_fixture)
    trainer.strategy = strategy_instance  # Patch the strategy with the fixture instance

    train_set, val_set = mnist_datasets
    trainer.execute(train_set, val_set)
``` 
I suppose it's debatable which one is more readable, and I'm not 100% it would work, but something to consider.

src/itwinai/loggers.py
```diff
@@ -480,6 +482,9 @@ def __init__(
         self.run_description = run_description
         self.run_name = run_name
 
+        if self.tracking_uri is None and os.environ.get('MLFLOW_TRACKING_URI'):
+            self.tracking_uri = os.environ.get('MLFLOW_TRACKING_URI')
+
         if self.tracking_uri is None:
             # Default MLFLow tracking URI
             saved_abs_path = os.path.abspath(self.savedir)
```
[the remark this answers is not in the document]
matbun: sick
added

src/itwinai/torch/distributed.py
```diff
@@ -41,29 +57,19 @@ class TorchDistributedStrategy(DistributedStrategy):
...
             bool: True if main worker.
         """
-        if not self.is_initialized:
-            raise UninitializedStrategyError(
-                "Strategy has not been initialized. Use the init method."
-            )
         return self.global_rank() == 0
 
     @abc.abstractmethod
     def init(self) -> None:
```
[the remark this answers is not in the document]
matbun: what do you mean by "default init"? It would be better to address this in a separate PR, though. Could you open an issue on this, please?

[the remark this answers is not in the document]
annaelisalappe: Yeah makes sense, I'll do that - I just mean the __init__ function that every class has, naming it differently would distance it more from that and make it clear that it is meant to be called after, not when the __init__ function is called

src/itwinai/torch/distributed.py
matbun: good find. Fixed it

tests/torch/distribtued_decorator.py
matbun: this is an old file not meant to be used by pytest. One day I should turn it into a test or drop the distributed decorator

tests/torch/runall.sh
matbun: You can change that path when you are using the script. Anyways, this script is for development only. I have added a disclaimer to the top

tests/torch/test_distribtued.py
matbun: why?

tests/torch/test_distribtued_training.py
matbun: this was deleted

env-files/torch/Dockerfile
matbun: added back. Thanks for noticing that

tests/torch/test_torch_trainer.py
matbun: interesting... I am using different markers for each test function in this case. Example:

```python
@pytest.mark.hpc
@pytest.mark.torch_dist
def test_distributed_trainer_ddp_mnist(mnist_datasets, ddp_strategy):
  ...

@pytest.mark.hpc
@pytest.mark.deepspeed_dist
def test_distributed_trainer_deepspeed_mnist(mnist_datasets, deepspeed_strategy):
  ...

@pytest.mark.hpc
@pytest.mark.horovod_dist
def test_distributed_trainer_horovod_mnist(mnist_datasets, horovod_strategy):
  ...
```

Do you know if specific markers could be added to the parametrization process somehow?

tests/torch/test_distribtued.py
annaelisalappe: Don't you mean test_distributed? Or was this on purpose?

src/itwinai/torch/distributed.py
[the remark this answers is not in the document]
matbun: ok, so definitely for another PR as it is going to require many changes

tests/torch/test_distribtued.py
annaelisalappe: True! It should be possible though: https://docs.pytest.org/en/stable/example/markers.html#marking-individual-tests-when-using-parametrize

matbun: ohh yes, I always make that typo I don't even recognize it yet

matbun: great find!

PR #237, opened 2024-10-28
Gpu monitoring

jarlsondre:
# Summary
Add decorator and CLI command for measuring GPU utilization. You can now add the decorator to any `execute()` function of a `TorchTrainer` class and it will log the files in a folder named `utilization_logs`. Then, when the training is finished, you can create a plot using the `itwinai generate-gpu-energy-plot`.

An example of a resulting plot (real data from training on EURAC) can be seen here: 
image

---

**Related issue :**
#221


src/itwinai/torch/monitoring/monitoring.py
```diff
@@ -0,0 +1,191 @@
...
+        log_df.to_csv(output_path, index=False)
+        print(f"Writing DataFrame to '{output_path}'.")
+
+    @functools.wraps(method)
+    def measured_method(self: TorchTrainer, *args, **kwargs) -> Any:
+        gpu_probing_interval = 1
+        warmup_time = 5
+
+        strategy = self.strategy
+        strategy.init()
```
matbun, line 117: Move strategy creation and destroy outside of the decorators, as it should not be their responsibility to manage that. In addition, it would make debugging more complicated.

src/itwinai/torch/monitoring/monitoring.py
```diff
@@ -0,0 +1,191 @@
...
+
+                # Converting the shared log to non-shared log
+                local_utilization_log = {key: list(data[key]) for key in data.keys()}
+                manager.shutdown()
+
+        global_utilization_log = strategy.gather_obj(local_utilization_log, dst_rank=0)
+        if strategy.is_main_worker:
+            write_logs_to_file(global_utilization_log, output_path)
+
+        strategy.clean_up()
```
matbun, line 188: Same as for init()

src/itwinai/torch/monitoring/monitoring.py
```diff
@@ -0,0 +1,191 @@
...
+        if isinstance(strategy, NonDistributedStrategy):
+            strategy_name = "non-dist"
+        elif isinstance(strategy, TorchDDPStrategy):
+            strategy_name = "ddp"
+        elif isinstance(strategy, DeepSpeedStrategy):
+            strategy_name = "deepspeed"
+        elif isinstance(strategy, HorovodStrategy):
+            strategy_name = "horovod"
+        else:
+            strategy_name = "unk"
```
matbun, line 128: You could also use `strategy.__class__.__name__` to do this in one line and automatically include future strategies

src/itwinai/torch/monitoring/monitoring.py
```diff
@@ -0,0 +1,191 @@
...
+
+        local_rank = strategy.local_rank()
+        global_rank = strategy.global_rank()
+        num_global_gpus = strategy.global_world_size()
+        num_local_gpus = torch.cuda.device_count()
```
matbun, line 133: why not using `strategy.local_world_size()`?

use-cases/eurac/trainer.py
```diff
@@ -92,7 +93,8 @@ def __init__(
         self.save_parameters(**self.locals2params(locals()))
 
     @suppress_workers_print
-    @profile_torch_trainer
+    # @profile_torch_trainer
```
matbun, line 96: Make sure that decorators can be used together (for future PRs, though)

src/itwinai/torch/monitoring/monitoring.py
jarlsondre: I took this opportunity to create a `name` field for the strategies, as I think that looks a bit cleaner. Besides, I think that the class names are a bit too verbose for logging purposes (IMO anyway).

use-cases/eurac/trainer.py
jarlsondre: I think that as soon as I fix the other issue with `init()` and `cleanup()`, this kinda solves itself 😎

use-cases/eurac/trainer.py
```diff
@@ -148,6 +148,7 @@ def set_epoch(self, epoch: int):
             self.train_loader.sampler.set_epoch(epoch)
             self.val_loader.sampler.set_epoch(epoch)
 
+    @measure_gpu_utilization
```
matbun, line 152: is it possible to use the profiler at the same time with the GPU utilization monitor?

use-cases/eurac/plots/comm_plot.png
matbun, line 1: Does it make sense to have only communication and no computation?

use-cases/eurac/trainer.py
jarlsondre: Yes

use-cases/eurac/plots/comm_plot.png
jarlsondre: In this particular example, I had trained the model for too few epochs for the profiler to have time to kick in. You have to train it for at least 4 epochs due to the values I set for the `wait` and `warmup` arguments to the profiler. I have tested this with more epochs and you will get a more reasonable plot in that case.

matbun, line 1: Is this plot still needed?

jarlsondre: Nope, deleted it now

annaelisalappe (approved):

matbun:
LGTM

matbun (approved):

PR #239, opened 2024-10-30
Scalability test wall clock

jarlsondre:
# Summary
Update the implementation for the original scalability test to include an absolute plot. In addition, many changes have been made to streamline the different scalability tests. 

Note: As I have been rebasing from the `gpu-monitoring` branch, many of the changes from that branch will appear in this PR until the other PR has been merged. 

**Related issue :**
#221


matbun (dismissed):
LGTM

matbun:
I just realized that it would be nice to enable/disable all scalability analyses from a flag passed to the Trainer's constructor. This would give the possibility to the user to decide when profiling should actually be done and when to have less overhead

jarlsondre:
> I just realized that it would be nice to enable/disable all scalability analyses from a flag passed to the Trainer's constructor. This would give the possibility to the user to decide when profiling should actually be done and when to have less overhead

The way I had it in mind was that the decorator was going to be used as a toggle for this. The obvious problem with doing that is of course that you have to actually change the source code every time you want to do it. However, if you add a flag to the constructor that overrides the decorator then you in a sense have two contradicting toggles, which to me sounds quite confusing. To counteract this, I would suggest then that the decorators are a permanent part of every trainer, somehow, so that we limit the number of places for configuration, as anything else quickly becomes convoluted.

matbun (approved):

annaelisalappe (approved):

PR #240, opened 2024-11-07
Virgo HDF5 file format

jarlsondre:
# Summary
Updates the data loading and synthetic data generation to use the HDF5 file format instead of pickle files. This makes the code both shorter and easier to work with. In particular, we are no longer limited to storing exactly 500 rows per file, but can rather choose a chunk size that suits us after the data has been created. Additionally, due to the functionality of the HDF5 format, all the data is stored in a single file without any issues with loading too much data into memory during runtime. 

**Related issue :**
None


use-cases/virgo/config.yaml
```diff
@@ -1,6 +1,9 @@
 root_folder: /p/scratch/intertwin/datasets/virgo
+hdf5_file_location: /p/project1/intertwin/saether1/itwinai/use-cases/virgo/data/virgo_data.hdf5
+hdf5_dataset_name: virgo_dataset
+chunk_size: 5
```
matbun, line 4: to avoid confusion, you could force the batch_size in the DataLoader to be always 1, allowing to control the actual batch size by means of the chunk_size. AFAIU, now the *real* batch size is equal to `batch_size  * chunk_size`. If this is the case, I would rename `chunk_size` in the top-level configuration as batch_size and remove batch_size. Down in the config I would then specify   

```yaml
          hdf5_file_location: ${hdf5_file_location}
          chunk_size: ${batch_size}
          hdf5_dataset_name: ${hdf5_dataset_name}
```

and where before we had  `batch_size: ${batch_size}` I would replace it with `batch_size: 1`.

See comments below

use-cases/virgo/config.yaml
```diff
@@ -17,6 +20,9 @@ training_pipeline:
           validation_proportion: 0.1
           rnd_seed: 42
           root_folder: ${root_folder}
+          hdf5_file_location: ${hdf5_file_location}
+          chunk_size: ${chunk_size}
```
matbun, line 24: I would possibly replace it as `chunk_size: ${batch_size}`

use-cases/virgo/config.yaml
```diff
@@ -17,6 +20,9 @@ training_pipeline:
...
+          hdf5_dataset_name: ${hdf5_dataset_name}
       - class_path: trainer.NoiseGeneratorTrainer
         init_args:
           config:
```
matbun, line 28: this would become

```yaml
config:
  generator: simple #unet
  batch_size: 1
  optim_lr: ${learning_rate}
  loss: l1
  save_best: true
  shuffle_train: true
```

use-cases/virgo/data.py
```diff
@@ -42,32 +40,103 @@ def execute(self) -> pd.DataFrame:
...
+        return normalize_(data)
+
+
 class SyntheticTimeSeriesDataset(Dataset):
-    def __init__(self, root_folder: Optional[str] = None,):
+    def __init__(
+        self,
+        root_folder: Optional[str] = None,
+    ):
         """Initialize the DataFrameDataset class.
```
matbun, line 140: add in the docstring that this dataset is deprecated

use-cases/virgo/synthetic_data_gen/file_gen.py
```diff
@@ -1,24 +1,53 @@
...
 import numpy as np
 import pandas as pd
+import h5py
 from gwpy.timeseries import TimeSeries
+from time import time
+from tqdm import tqdm
 
-from ..src.dataset import generate_cut_image_dataset
+# sys.path.append(str(Path("..").resolve()))
+sys.path.append(str(Path.cwd().resolve()))
```
matbun, line 13: interesting. What is this doing?

use-cases/virgo/synthetic_data_gen/file_gen.py
```diff
@@ -1,24 +1,53 @@
...
+        raise ValueError(
+            f"'array' has an incorrect shape: ({actual_shape_str}). "
+            f"Should have been (x, {expected_shape_str})."
+        )
+
+    print(f"Appending to file: '{str(file_path.resolve())}'.")
+    with h5py.File(file_path, "a") as f:
+        dset = f[dataset_name]
+        dset.resize(dset.shape[0] + array.shape[0], axis=0)
+        dset[-array.shape[0] :] = array
```
matbun, line 36: is this thread safe? Perhaps HDF5 is ensuring it somehow?

jarlsondre: I doubt it's thread safe, but I am using a single file per process so I don't think it should be a problem.

.github/workflows/lint.yml
annaelisalappe, line 1: This is more of a general question, but why is it necessary for us to use super-linter? Should ruff not be able to do all the linting for us by itself?

use-cases/virgo/data.py
```diff
@@ -42,98 +51,97 @@ def execute(self) -> pd.DataFrame:
...
         return df
 
 
-class SyntheticTimeSeriesDataset(Dataset):
-    def __init__(self, root_folder: Optional[str] = None,):
+class SyntheticTimeSeriesDatasetHDF5(Dataset):
+    def __init__(
+        self,
+        hdf5_file_location: str,
+        chunk_size: int = 500,
```
annaelisalappe, line 85: I don't really understand the purpose of the chunk_size argument, what is that needed for?

use-cases/virgo/synthetic-data-gen/data_generation_hdf5.sh
```diff
@@ -0,0 +1,27 @@
...
+#SBATCH --partition=develbooster
+#SBATCH --array=1-75
+#SBATCH --job-name=generate_virgo_data
+#SBATCH --cpus-per-task=26
+
+# Load required modules
+ml Stages/2024 GCC OpenMPI CUDA/12 MPI-settings/CUDA Python HDF5 PnetCDF libaio mpi4py
+
+# Activate Python virtual environment
+source ../../envAI_hdfml/bin/activate
```
annaelisalappe, line 17: Maybe change this to the jewels environment (../../envAI_juwels/bin/activate). This will come up a few more times probably

use-cases/virgo/synthetic-data-gen/file_gen_hdf5.py
```diff
@@ -0,0 +1,250 @@
...
+import sys
+from time import time
+
+import h5py
+import numpy as np
+import pandas as pd
+from gwpy.timeseries import TimeSeries
+from tqdm import tqdm
+
+# sys.path.append(str(Path("..").resolve()))
```
annaelisalappe, line 21: Can this be removed?

.github/workflows/lint.yml
matbun: super linter is linting other files (e.g., Dockerfile) which I guess ruff is not able to

use-cases/virgo/config.yaml
```diff
@@ -1,7 +1,8 @@
-root_folder: /p/scratch/intertwin/datasets/virgo
+hdf5_file_location: /p/scratch/intertwin/datasets/virgo_hdf5/virgo_data.hdf5
+hdf5_dataset_name: virgo_dataset
 data_root: ./data
-epochs: 2
-batch_size: 3
+epochs: 5
+batch_size: 2000 # equivalent to chunk size
```
matbun, line 5: are you sure that you can comment inline?

uv.lock
matbun, line 1: maybe a bit premature to add this before extensive tests with uv?

use-cases/virgo/config.yaml
jarlsondre: As in whether it crashes or not or as in whether it is best practice?

use-cases/virgo/synthetic-data-gen/data_generation_hdf5.sh
jarlsondre: I agree that this should be changed. I have it this way right now since I haven't changed the name of my venv, but I think I might have a better idea about how to handle this naming thing anyway. See Mattermost.

.github/workflows/lint.yml
jarlsondre: Yeah Ruff is a python-only linter and thus wouldn't work for other linting purposes, while the super-linter can essentially lint anything, e.g. yaml, c++, markdown and most other things. Even though we don't use that many linters atm, it would make it easy to add new ones in the future and it doesn't really cost us anything to keep it the way it is, other than potentially having two linting configs for each linter (which is kinda annoying, I suppose).

use-cases/virgo/synthetic-data-gen/file_gen_hdf5.py
jarlsondre: I think so, yes, but I was kind of unsure about which one to keep, haha 😅

Since this is just a script anyway, it probably doesn't matter that much 🤗

uv.lock
jarlsondre: Haha yes that was a mistake

use-cases/virgo/data.py
annaelisalappe: Ah okay, then it's what I thought initially and pretty intuitive, I think it doesn't need explaining then :)

use-cases/virgo/config.yaml
matbun: whether it crashes. That's sth you cannot do with a workflow files for GH actions, as it will consider the whole line as a string, inclluding the comment (as far as I remember). In this case the YAML parser may be smarter though, as it expects an int.

In general it does not look like a good practice, although I may also have been doing it in these files.

use-cases/virgo/synthetic_data_gen/file_gen.py
matbun: I see, thanks

annaelisalappe (dismissed):

annaelisalappe:
Looks good I think!

annaelisalappe (approved):

matbun (approved):

PR #241, opened 2024-11-08
Lighting logger integration

annaelisalappe:
This PR adds an Adapter (Logger) to enable logging from within a Lightning Trainer. This is necessary because Lightning provides its own logger component, which any custom loggers have to comply with. However, should a user want to log using an itwinai logger e.g. the Prov4MLLogger, they need an adapter to convert the lightning logger calls to calls to the itwinai logger.


src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+    @rank_zero_experiment
+    def experiment(self) -> Logger:
+        """Lightning Logger function.
+        Initializes and returns the itwinai Logger context for experiment tracking.
+
+        Returns:
+            Logger: The itwinai logger instance.
+        """
+        if not self._initialized:
+            # With the rank_zero_decorator the rank will always be 0
```
matbun, line 85: I guess you mean "rank_zero_experiment decorator" instead of "rank_zero_decorator"

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+                # Log the aliases
+                with open(f"{tmp_dir}/aliases.txt", "w") as tmp_file_aliases:
+                    tmp_file_aliases.write(str(aliases))
+
+                # Log the metadata and aliases
+                self.experiment.log(
+                    item=tmp_dir,
+                    identifier=artifact_path,
+                    kind='artifact'
+                )
```
matbun, line 236: Are you sure that by passing the path to the directory all the files in it will be logged as artifacts? Is this working for all logging frameworks? If I well remember, by default they expect the path to a file. For instance MLFlow has a specific function ([`log_artifacts`](https://mlflow.org/docs/latest/python_api/mlflow.html#mlflow.log_artifacts)) to log all the artifacts in a folder.

tests/torch/test_loggers.py
matbun, line 1: there is another file with this name, which will make pytest fail when collecting tests. I'd suggest to rename it.

The unit tests seem OK, but I would suggest adding some integration tests both with the itwinai loggers and with the lightning module and trainer. You can create a set of simple tests with minimal training and validation using a lightning module, trainer and itwinai loggers (e.g., wrapped in a `LoggersCollection`)

tests/torch/test_loggers.py
```diff
@@ -0,0 +1,111 @@
...
+
+    return itwinai_logger_mock, lightning_logger
+
+
+def test_lightning_experiment_property(lightning_loggers):
+    """Test that the experiment property initializes
+    the logger context and returns the logger."""
+    itwinai_logger_mock, lightning_logger = lightning_loggers
+    experiment = lightning_logger.experiment
+    assert isinstance(experiment, Logger)
```
matbun, line 34: Have you tested this locally? I think this will hardly pass being experiment a plain `MagicMock`

tests/torch/test_loggers.py
```diff
@@ -0,0 +1,111 @@
+# --------------------------------------------------------------------------------------
+# Part of the interTwin Project: https://www.intertwin.eu/
+#
+# Created by: Anna Elisa Lappe
+#
+# Credit:
+# - Anna Lappe <anna.elisa.lappe@cern.ch> - CERN
```
matbun, line 7: Could you add me below here pls?

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
+# --------------------------------------------------------------------------------------
+# Part of the interTwin Project: https://www.intertwin.eu/
+#
+# Created by: Anna Elisa Lappe
+#
+# Credit:
+# - Anna Lappe <anna.elisa.lappe@cern.ch> - CERN
```
matbun, line 7: Could you add me below here pls?

tests/torch/test_loggers.py
```diff
@@ -0,0 +1,111 @@
...
+    for key, value in metrics.items():
+        itwinai_logger_mock.log.assert_any_call(
+            item=value, identifier=key, kind='metric', step=step)
+
+
+def test_lightning_log_hyperparams(lightning_loggers):
+    """Test log_hyperparams method to ensure it logs hyperparameters correctly."""
+    itwinai_logger_mock, lightning_logger = lightning_loggers
+    dict_params = {"learning_rate": 0.001, "batch_size": 32}
+    namespace_params = Namespace(**dict_params)
```
matbun, line 53: to avoid confusion I would move this line below before `namespace_params` is actually used

tests/torch/test_loggers.py
```diff
@@ -0,0 +1,111 @@
...
+from lightning.pytorch.callbacks.model_checkpoint import ModelCheckpoint
+
+from itwinai.loggers import Logger
+from itwinai.torch.loggers import PyTorchLightningLogger
+
+
+@pytest.fixture
+def lightning_loggers():
+    """Setup a generic PyTorchLightningLogger with mock loggers for testing."""
+    itwinai_logger_mock = MagicMock()
```
matbun, line 23: Maybe use `itwinai_logger_mock = MagicMock(spec=Logger)` ?

matbun (changes requested):
The lightning's logger for itwinai seems very well done. I would improve a couple of things I left as comments and add more tests. Also, the CI pipeline is currently failing since there are two test modules with the same name (see comments)

Could you add the loggers module here? https://github.com/interTwin-eu/itwinai/blob/86f536f57dfaa9ea9ad433f39c152ee80cfccd36/src/itwinai/tests/sanity_check.py#L25

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+            self._initialized = True
+
+        return self.itwinai_logger
+
+    @override
+    @rank_zero_only
+    def finalize(
+        self,
+        status: str
+    ) -> None:
```
jarlsondre, line 96: This can be a single line, i.e. `def finalize(self, status: str) -> None`

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+            self.itwinai_logger.create_logger_context(rank=0)
+            self._initialized = True
+
+        return self.itwinai_logger
+
+    @override
+    @rank_zero_only
+    def finalize(
+        self,
+        status: str
```
jarlsondre, line 95: I can see that `status` is unused, but you need it because it is also defined in the super class' signature. I would note this in the docstring, maybe even instead of what it currently says, and potentially set a default argument so that you don't have to pass it in. Not sure how well that works with other functions calling it, but should be fine?

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+            metrics: Dict[str, float],
+            step: Optional[int] = None
+    ) -> None:
+        """Lightning Logger function.
+        Logs the given metrics and is usually called by the Lightning Trainer.
+
+        Args:
+            metrics (Dict[str, float]): Dictionary of metrics to log.
+            step (Optional[int], optional): Training step associated with the metrics.
+            Defaults to None.
```
jarlsondre, line 125: indent "defaults to none" :)

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+                kind='metric',
+                step=step
+            )
+
+    @override
+    @rank_zero_only
+    def log_hyperparams(
+            self,
+            params: Union[Dict[str, Any], Namespace]
+    ) -> None:
```
jarlsondre, line 140: This fits in a single line

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+        if isinstance(params, Namespace):
+            params = vars(params)
+
+        self.experiment.save_hyperparameters(params)
+
+    @override
+    def after_save_checkpoint(
+        self,
+        checkpoint_callback: ModelCheckpoint
+    ) -> None:
```
jarlsondre, line 155: This fits in a single line

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+        The checkpoints are logged as artifacts.
+
+        Args:
+            checkpoint_callback (ModelCheckpoint): Callback instance to manage checkpointing.
+        """
+        if self._log_model == "all" or (
+            self._log_model is True and checkpoint_callback.save_top_k == -1
+        ):
+            self._scan_and_log_checkpoints(checkpoint_callback)
+        elif self._log_model is True:
```
jarlsondre, line 167: I like the use of "is" here, I would not even have thought of that haha.

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+        parameters.
+
+        Args:
+            checkpoint_callback (ModelCheckpoint): Callback instance used for checkpointing.
+        """
+        # Get checkpoints to be saved with associated score
+        checkpoints = _scan_checkpoints(checkpoint_callback, self._logged_model_time)
+
+        # Log iteratively all new checkpoints
+        for t, p, s, tag in checkpoints:
```
jarlsondre, line 187: I feel like I recognize this code from something I saw in the Lightning docs, so that might be the reason, but I think you are much better off here naming the variables something else than "t", "p" etc., as I'm not sure what they mean atm

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+            else:
+                aliases = ["latest"]
+            artifact_path = Path(p).stem
+
+            # Log the checkpoint
+            self.experiment.log(
+                item=p,
+                identifier=artifact_path,
+                kind='artifact'
+            )
```
jarlsondre, line 216: This fits in a single line

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+                        "_every_n_train_steps",
+                    ]
+                    # ensure it does not break if `ModelCheckpoint` args change
+                    if hasattr(checkpoint_callback, k)
+                },
+            }
+            if p == checkpoint_callback.best_model_path:
+                aliases = ["latest", "best"]
+            else:
+                aliases = ["latest"]
```
jarlsondre, line 208: I think this could be reduced to 

```python
aliases = ["latest"]
if p == ...
    aliases.append("best")
```
This would save you a line, but I suppose one can argue which is more readable 🤗

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+                item=p,
+                identifier=artifact_path,
+                kind='artifact'
+            )
+
+            with tempfile.TemporaryDirectory(
+                prefix="test",
+                suffix="test",
+                dir=os.getcwd()
+            ) as tmp_dir:
```
jarlsondre, line 222: In this case it would probably be nicer to do 
```python
with ...(
   prefix, suffix, dir
) as tmp_dir:
```
i.e. that you have the middle ones on a single line.

tests/torch/test_loggers.py
```diff
@@ -0,0 +1,111 @@
...
+    lightning_logger = PyTorchLightningLogger(itwinai_logger=itwinai_logger_mock)
+
+    return itwinai_logger_mock, lightning_logger
+
+
+def test_lightning_experiment_property(lightning_loggers):
```
jarlsondre, line 29: I have some trouble understanding exactly what this is testing. The function that is being tested has at most three lines of actions, two under the if statement and one return statement. This test checks that the output type is correct and that one of the functions inside the if statements is being called. 

In other words, this test really just seems to be replicating the actual implementation of the code, which doesn't really offer much value beyond the original code itself. If you wanted to take this one step further, you could check that "self._initialized == True", at which point the test would be equivalent to the actual implementation itself. This means that if you ever wanna change the implementation but leave the signature the same, you would have to rewrite the tests even if the function does what it's supposed to.

tests/torch/test_loggers.py
```diff
@@ -0,0 +1,111 @@
...
+    itwinai_logger_mock.save_hyperparameters.assert_called_once_with(dict_params)
+
+    itwinai_logger_mock.reset_mock()
+    lightning_logger.log_hyperparams(params=namespace_params)
+    itwinai_logger_mock.save_hyperparameters.assert_called_once_with(dict_params)
+
+
+def test_lightning_after_save_checkpoint(lightning_loggers):
+    """Test after_save_checkpoint method in PyTorchLightningLogger."""
+    itwinai_logger_mock, lightning_logger = lightning_loggers
```
jarlsondre, line 65: unused variable, would rather do`_ , lightning_logger = ...`

tests/torch/test_loggers.py
```diff
@@ -0,0 +1,111 @@
...
+
+    checkpoint_callback = MagicMock(spec=ModelCheckpoint)
+    checkpoint_callback.save_top_k = -1
+    checkpoint_callback.best_model_path = "/path/to/best_model.ckpt"
```
jarlsondre, line 69: Would be better to use pathlib for this path

matbun:
@annaelisalappe it seems that you have a doppelganger in the commits history

Also, could you update this page with an example on the lightning logger please? https://itwinai.readthedocs.io/latest/how-it-works/loggers/explain_loggers.html

src/itwinai/torch/loggers.py
annaelisalappe: Hahaha

annaelisalappe: I think this is one of the weird habits I have because I find it nicer, but probably not exactly best practice

src/itwinai/torch/loggers.py
annaelisalappe: Good point, it actually took me so long to understand this myself

src/itwinai/torch/loggers.py
annaelisalappe: Actually though, because _log_model could also be a string, I think this makes it more clear what we are testing for, does that make sense?

tests/torch/test_loggers.py
annaelisalappe: Hm so you mean you would suggest removing this test? Maybe I can add a shorter version of this to the function to test the .log call, as there the experiment is also initialised

src/itwinai/torch/loggers.py
matbun: I think that `is True` makes the conditional more robust to misuses of user input. Imagine if the user is passing `log_models=42` or `log_models="hello"`... In both cases the control flow could be hijacked.

It seems that this is the same way they used in the MLFlow logger implementation: https://github.com/Lightning-AI/pytorch-lightning/blob/3627c5bfac704d44c0d055a2cdf6f3f9e3f9e8c1/src/lightning/pytorch/loggers/mlflow.py#L328

jarlsondre: Yeah in this case I specifically liked it because of Lightning's bad choices regarding using just a single variable to control what seems like multiple things. The "is" works as a nice fix for that.

tests/torch/test_loggers.py
jarlsondre: To be honest, I am not entirely sure. It might be useful, but when comparing it to the actual function implementation I feel like they are so similar that I am not sure what the test adds. That being said, I am unfamiliar with Lightning so I don't think I'd be able to fully assess the value of the test. It's more meant as "food for thought" and a way for me to grasp more of the domain than anything else, really :D

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+
+        self._initialized = False
+
+    @property
+    def name(self) -> Optional[str]:
+        """Return the experiment name."""
+
+    @property
+    def version(self) -> Optional[Union[int, str]]:
+        """Return the experiment version."""
```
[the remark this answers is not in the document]
annaelisalappe: I have incorporated this somewhat minimally for now, I would look more closely at what values make sense for the WandB Logger, Tensorboard Logger etc, when I refactor the loggers to see which variables can be consolidated etc.

tests/torch/test_loggers.py
annaelisalappe: Good catch! I have removed this function because, as jarl pointed out, it didn't add very much value, and I think what it is testing is covered sufficiently by other functions. I did add the `spec=` argument to all other MagicMocks though, that is definitely better practice!

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+import yaml
+from lightning.pytorch.callbacks.model_checkpoint import ModelCheckpoint
+from lightning.pytorch.loggers.logger import Logger as LightningLogger
+from lightning.pytorch.loggers.logger import rank_zero_experiment
+from lightning.pytorch.loggers.utilities import _scan_checkpoints
+from lightning.pytorch.utilities import rank_zero_only
+from torch import Tensor
+from typing_extensions import override
+
+from itwinai.loggers import Logger
```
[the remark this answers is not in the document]
annaelisalappe: Added!

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+class PyTorchLightningLogger(LightningLogger):
```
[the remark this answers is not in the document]
annaelisalappe: Added!

src/itwinai/torch/loggers.py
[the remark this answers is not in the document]
matbun: ok great!

tests/torch/test_loggers.py
annaelisalappe: Done!

src/itwinai/loggers.py
```diff
@@ -1,3 +1,13 @@
+# --------------------------------------------------------------------------------------
+# Part of the interTwin Project: https://www.intertwin.eu/
+#
+# Created by: Anna Elisa Lappe
```
matbun, line 4: I think I created it :)

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,222 @@
...
+            aliases = (
+                ["latest", "best"]
+                if path == checkpoint_callback.best_model_path
+                else ["latest"]
+            )
+
+            artifact_path = Path(path).stem
+
+            print(f"Path: {path}")
+            print(f"Identifier: {artifact_path}")
```
matbun, line 201: I would remove these prints. Or is there a reason to have them?

tests/loggers/test_loggers.py
```diff
@@ -0,0 +1,152 @@
+# --------------------------------------------------------------------------------------
+# Part of the interTwin Project: https://www.intertwin.eu/
+#
+# Created by: Anna Elisa Lappe
```
matbun, line 4: If I well remember I created this. If I am right, you can anyways add yourself to the credit ;)

annalappe: Omg haha sorry, copy paste got me 🙈

src/itwinai/torch/loggers.py
annalappe: Nope, I will remove them :)

annaelisalappe: Removed them

src/itwinai/loggers.py
```diff
@@ -350,13 +374,18 @@ def create_logger_context(self, rank: Optional[int] = None):
...
-        run_dirs = sorted([int(dir) for dir in os.listdir(self.savedir)])
-        if len(run_dirs) == 0:
-            self.run_id = 0
+        # os.makedirs(self.savedir, exist_ok=True)
+        if os.path.isdir(self.savedir):
+            numeric_dirs = [
+                int(name)
+                for name in os.listdir(self.savedir)
+                if os.path.isdir(os.path.join(self.savedir, name)) and name.isdigit()
+            ]
```
jarlsondre, line 383: I think this section would be a bit cleaner with pathlib, e.g. if `self.savedir` is a Path object then you can just do `(self.savedir / name).is_dir()` which is considerably shorter.

src/itwinai/loggers.py
```diff
@@ -400,25 +429,40 @@ def log(
...
+                    int(name)
+                    for name in os.listdir(self.savedir)
+                    if os.path.isdir(os.path.join(self.savedir, name))
+                    and name.isdigit()
+                ]
+                child_id = max(numeric_dirs) + 1
+                target_path = os.path.join(
+                    artifact_dir, f"{self.experiment_id}.{child_id}"
+                )
+                shutil.copytree(item, target_path, dirs_exist_ok=True)
```
jarlsondre, line 452: more pathlib stuff

src/itwinai/loggers.py
```diff
@@ -400,25 +429,40 @@ def log(
...
+                        The item will be ignored and not logged.")
+
         elif kind == "torch":
-            identifier = os.path.join(self.run_path, identifier)
-            print(f"ConsoleLogger: Saving to {identifier}...")
             import torch
 
-            torch.save(item, identifier)
-        else:
+            target_path = os.path.join(self.run_path, identifier)
```
jarlsondre, line 461: pathlib

src/itwinai/loggers.py
```diff
@@ -477,14 +521,14 @@ def __init__(
...
         self.tracking_uri = tracking_uri
         self.run_description = run_description
         self.run_name = run_name
+        self.experiment_name = experiment_name
 
         self.tracking_uri = (
             self.tracking_uri
-            or os.environ.get('MLFLOW_TRACKING_URI')
+            or os.environ.get("MLFLOW_TRACKING_URI")
             or pathlib.Path(os.path.abspath(self.savedir)).as_uri()
```
jarlsondre, line 532: I know this isn't code you changed but this hurts a bit because one can just do `.resolve()` instead of `os.path.abspath`

src/itwinai/torch/loggers.py
```diff
@@ -0,0 +1,219 @@
...
+        """Lightning Logger function.
+        Logs the given metrics and is usually called by the Lightning Trainer.
+
+        Args:
+            metrics (Dict[str, float]): Dictionary of metrics to log.
+            step (Optional[int], optional): Training step associated with the metrics.
+                Defaults to None.
+        """
+        for key, value in metrics.items():
+            self.experiment.log(item=value, identifier=key, kind="metric", step=step)
```
jarlsondre, line 130: This is a very small thing and I am not even sure about it, but it could be nicer name the variables differently than "key" and "value" here, since you are using them for something else. E.g. you could say
```
for identifier, item in metrics.items(): 
```
That being said, I am not super opposed to the idea of calling them "key" and "value" since it is a dictionary, but I think descriptive names are always better. Really not that important in this two-line example though, haha, so you can leave it if you want.

src/itwinai/loggers.py
annaelisalappe: Added

src/itwinai/loggers.py
```diff
@@ -140,10 +150,24 @@ def __init__(
...
+
+    @property
+    def experiment_id(self) -> Optional[str]:
+        """Return the experiment name."""
+        return self._experiment_id
+
+    @property
+    def run_id(self) -> Optional[Union[int, str]]:
+        """Return the experiment version."""
+        return self._run_id
```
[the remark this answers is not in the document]
annaelisalappe: This was inspired by the way it is done in the lightning logger - honestly though I have always wondered this about the @property decorator anyway. The idea was also to refactor the loggers in the future, and then we would consolidate some of the arguments to the loggers and try to have one experiment_id and run_id in each logger, that should hopefully be enough to characterise that logger instance.

matbun (approved):

Issue #244, opened 2024-11-11
Update package manager to uv


jarlsondre:
# Update the package manager to uv
[uv](https://docs.astral.sh/uv/) is a new package manager written in Rust, which is quite a bit faster than other alternatives. It is also meant to handle more things than just installing packages, such as creating python environments. Currently, our install scripts are quite messy and take a long time to run, so updating the scripts and making them conform to `uv`'s standard would clean up our code quite a bit. In addition to this, it would most likely be a lot faster. 

## Suggested solution
Rewrite the scripts such as `generic_torch.sh` and incorporate all of this into the `pyproject.toml` for a more unified experience. Make sure that we are not dependent on `uv`, so we also have to use files that can be used with `pip`.

PR #245, opened 2024-11-13
Venv creation and uv support

jarlsondre:
# Summary
This PR updates the way we create venvs removing parts of the reliance on massive scripts such as `generic_torch.sh`. It also allows us to use `uv`, which is quite a bit faster than just using `pip`. Of course, this is done without depending on `uv`, as it is still possible to only use `pip`. 

# Motivation
The reliance on the `generic_torch.sh` script has numerous disadvantages. First of all, we are unable to provide all of our dependencies in the `pyproject.toml`, which means that a simple `pip install itwinai` is simply not possible at the moment. Secondly, the `generic_torch.sh` script is messy with many if statement that are repeated multiple times throughout, such as "if cuda" (but in shell syntax), meaning that it is hard to build upon the script. Thirdly, because we have a bunch of separated `pip install x` statements, we effectively give no way for the dependency manager to solve our dependency graph in a nice way. This results in many packages being installed only to be uninstalled on the next `pip install` statement. This causes our script to take much longer than needed. 

# Noteworthy
Because part of this PR is also about transitioning to `uv`, I have renamed a lot of the venvs that we use to just `.venv`. It seems that this work better with `uv` in some cases. I was also thinking that we could keep our old venvs and just symlink them to `.venv`, so that we don't have to create new names if we ever have to change systems again. This should hopefully streamline our naming convention a bit more. 

**Related issue :**
#244


uv-tutorial.md
```diff
@@ -0,0 +1,98 @@
...
+To add a new package to the project (i.e. to the `pyproject.toml` file) with `uv`, you
+can use the following command:
+
+```bash
+uv add <package>
+```
+
+> [!NOTE]
+> This will add the package to your `.venv` venv, so make sure to have symlinked to
+> this directory if you haven't already.
```
annaelisalappe, line 98: Looks so good! I only have one question - how does it work if we need to have multiple environments, such as one for Vega and one for Juwels? How would that work with the .venv / symlinks?

jarlsondre: This is an important point. 

1) I don't think it's a problem for Vega vs. Juwels since we don't push our environments and we would never deliberately make a Juwels environment on Vega and vice-versa. 
2) This is however a potential problem if you have one venv for TensorFlow and one for PyTorch. In this case, you would simply have to swap the symlinks when you needed to. If you have to run both at the same time then you would have to hardcode one of them (which is what we are doing now, by the way) to its specific venv.

annaelisalappe: Okay, that sounds reasonable I would say. Maybe add this to the tutorial, where you mention the symlink stuff?

pyproject.toml
```diff
@@ -69,6 +67,12 @@ dev = [
     "isort",
     "tensorflow==2.16.*",  # needed by tests on tensorboard
```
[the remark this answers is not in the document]
jarlsondre: Let me know if you think this fits the scope of this PR or if that should be something to work on later.

matbun:
If this PR is merged after https://github.com/interTwin-eu/itwinai/pull/249, we need to update the pyproject.toml to use the `new-main` version of the provenance logger

uv-tutorial.md
```diff
@@ -0,0 +1,98 @@
...
+```bash
+uv sync --extra torch --extra macos
+```
+
+These will usually correspond to the optional dependencies in the `pyproject.toml`. In
+particular, if you are a developer you would use one of the following two commands. If
+you are on HPC with cuda, you would use:
+
+```bash
+uv sync --extra torch --extra dev --extra linux \
```
matbun, line 64: need to remove linux

uv-tutorial.md
```diff
@@ -0,0 +1,98 @@
...
+
+```bash
+uv sync
+```
+
+If the `uv.lock` file has optional dependencies (e.g. `macos` or `torch`), then these
+can be added with the `--extra` flag as follows:
+
+```bash
+uv sync --extra torch --extra macos
```
matbun, line 56: This is cool. So basically we can have a single lock file which locks both macos and linux dependencies in the same place, right?

uv-tutorial.md
```diff
@@ -0,0 +1,98 @@
...
+## Adding new packages to the project
+
+To add a new package to the project (i.e. to the `pyproject.toml` file) with `uv`, you
+can use the following command:
+
+```bash
+uv add <package>
+```
+
+> [!NOTE]
```
matbun, line 96: maybe make this a warning

use-cases/eurac/requirements.txt
```diff
@@ -4,3 +4,4 @@ tqdm
 cf_xarray
 requests
 aiohttp
+ray
```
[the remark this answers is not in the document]
jarlsondre: It actually isn't (in the main branch anyway?), but adding it might be good.

src/itwinai/torch/distributed.py
```diff
@@ -400,7 +400,9 @@ def init(self) -> None:
                 which is already initialized.
         """
         if not distributed_resources_available():
-            raise RuntimeError("Trying to run distributed on insufficient resources.")
+            raise RuntimeError(
+                "Trying to run distributed on insufficient resources."
+            )
```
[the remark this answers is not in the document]
jarlsondre: Yeah I think it wasn't picking up the settings properly, but I think it might be better now. Probably defaulted to a shorter line length or something.

uv-tutorial.md
jarlsondre: That is how I understand it, yeah 😎

use-cases/eurac/requirements.txt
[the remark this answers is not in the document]
matbun: as of now it gets installed in the generic script. We can add it in the pyproject toml

uv-tutorial.md
matbun: cool

matbun (approved):

PR #238, opened 2024-10-29
Update contributors

matbun:
Update contributors in the docs

PR #246, opened 2024-11-14
Update README.md

r-sarma:
# Summary

---

**Related issue :**


matbun:
We already have another PR on which we are _collaboratively_ discussing this: https://github.com/interTwin-eu/itwinai/pull/238/
So please close this to avoid unnecessary duplication

PR #253, opened 2024-11-20
Update docs for use cases with EURAC

jarlsondre:
# Summary
Add EURAC to the list of use cases found in the docs and change the use-cases page slightly for a (IMO) cleaner look.

# Noteworthy
I chose to not add more than the readme for EURAC as I found that adding the code files is messy and doesn't add anything. Personally, I feel that if someone wants to explore the code, they can either do it on GitHub or (even better) in their own text editor.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -55,6 +55,7 @@ contains thoroughly tested features aligned with the toolkit's most recent relea
 
    getting-started/getting_started_with_itwinai
    getting-started/slurm
+   getting-started/plugins
 
 .. toctree::
    :maxdepth: 2
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -55,7 +55,6 @@ contains thoroughly tested features aligned with the toolkit's most recent relea
 
    getting-started/getting_started_with_itwinai
    getting-started/slurm
-   getting-started/plugins
 
 .. toctree::
    :maxdepth: 2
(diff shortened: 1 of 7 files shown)


docs/index.rst
```diff
@@ -55,7 +55,6 @@ contains thoroughly tested features aligned with the toolkit's most recent relea
 
    getting-started/getting_started_with_itwinai
    getting-started/slurm
-   getting-started/plugins
```
matbun, line 58: :(

matbun:
> I chose to not add more than the readme for EURAC as I found that adding the code files is messy and doesn't add anything. Personally, I feel that if someone wants to explore the code, they can either do it on GitHub or (even better) in their own text editor.

I agree, could you clean that up for the other use cases as well, please? Still, it would be nice to have a link to the repo's directory of the use case, for each use case, if possible

jarlsondre:
> > I chose to not add more than the readme for EURAC as I found that adding the code files is messy and doesn't add anything. Personally, I feel that if someone wants to explore the code, they can either do it on GitHub or (even better) in their own text editor.
> 
> I agree, could you clean that up for the other use cases as well, please? Still, it would be nice to have a link to the repo's directory of the use case, for each use case, if possible

I agree with adding the GitHub link. I actually thought I wrote that, but I see now that I only wrote it in my personal notes, hahaha

docs/index.rst
matbun: you should clone with `--recurse-submodules`

matbun (changes requested):
Apart from the plugins which should be added back, it looks good to me

matbun (approved):

PR #254, opened 2024-11-20
Ray torchtrainer integration

annaelisalappe:
# Summary
This PR adds the `RayTorchTrainer` class, as well as a `RayDistributedStrategy`, of which two are implemented so far (DDP and DeepSpeed). The RayTorchTrainer allows us to perform distributed hyperparameter optimisation, parallelising both the trials and the training within the trials themselves. I also implemented a new `RayNoiseGeneratorTrainer` class for the Virgo use case, to demonstrate how the RayTorchTrainer would be used. This integration is still a work in in progress, so this PR serves more as an intermediary consolidation, rather than a fully finished integration. Some things that are already on my mind to include/ improve in the next step are:

- Adding tests and tutorials to document the new RayTorchTrainer class and Strategies
- Adding further type checks to the different configs, as well as changing the current dictionary setup to something using a class that implements the `__get_item()__` method
- - Extending the possible search algorithms and schedulers, as well as adding more checks and warnings to ensure all required fields are correctly set 
- Adding more distributed strategies
 
However, what should work with this addition already is the execution of the NoiseGeneratorTrainer with Virgo, using two new pipelines defined in `use-cases/virgo/config.yaml`. These pipelines should be executable with both ddp and deepspeed strategies and with mostly any settings for the tune/scaling config, such as distributing individual trials across multiple nodes, using different search algorithms and schedulers, and so on).


README.md
```diff
@@ -59,8 +59,8 @@ environment for PyTorch:
 
     ```bash
     ml --force purge
-    ml Python CMake/3.24.3-GCCcore-11.3.0 mpi4py OpenMPI CUDA/11.7
-    ml GCCcore/11.3.0 NCCL/2.12.12-GCCcore-11.3.0-CUDA-11.7.0 cuDNN
+    ml Python CMake/3.24.3-GCCcore-11.3.0 mpi4py OpenMPI CUDA/12.3
```
jarlsondre, line 62: One thing I have been thinking about is whether we should enforce `Python/3.11.3` instead of just `Python` here. This is not directly related to your PR, but I would like your opinion on changing this across the code base. We could do this in a different branch or something, in which case. Doing it here would also be fine though.

src/itwinai/torch/distributed.py
```diff
@@ -3,10 +3,14 @@
 import os
 from typing import Any, Callable, Iterable, List, Literal, Optional, Tuple, Union
 
+import deepspeed
```
jarlsondre, line 6: Didn't we talk about putting these imports inside of the place where we use them, i.e. the relevant strategies?

annaelisalappe: True! I moved it down into the `init()` function of the `RayDeepSpeedStrategy`

src/itwinai/torch/distributed.py
```diff
@@ -1010,3 +1012,177 @@ def gather(self, tensor: torch.Tensor, dst_rank: int = 0):
...
+            head_node_ip = os.environ.get("head_node_ip")
+
+            if not ip_head or not head_node_ip:
+                raise EnvironmentError(
+                    "Ray initialization requires 'ip_head' and 'head_node_ip' to be set."
+                )
+
+            if not ray.is_initialized():
+                ray.init(address="auto")
+                print("Ray initialized!")
```
jarlsondre, line 1030: Not a big deal, but you could do the check `is_initialized()` at the beginning and return from the function if it isn't.

annaelisalappe: Added!

src/itwinai/torch/distributed.py
```diff
@@ -1010,3 +1012,177 @@ def gather(self, tensor: torch.Tensor, dst_rank: int = 0):
...
+    def local_rank(self) -> int:
+        return ray.train.get_context().get_local_rank()
+
+    @check_initialized
+    def distributed(
+        self,
+        model: nn.Module,
+        optimizer: Optimizer,
+        lr_scheduler: Optional[LRScheduler] = None,
+    ) -> Tuple[Module | Optimizer | LRScheduler | None]:
```
jarlsondre, line 1065: This should probably be: 

```
Tuple[Module, Optimizer, LRScheduler | None]
```
right?

use-cases/virgo/trainer.py
```diff
@@ -373,7 +370,279 @@ def train(self):
...
+            init_weights(self.model, "normal", scaling=0.02)
+        else:
+            raise ValueError("Unrecognized generator type! Got", generator)
+
+        # Select loss
+        loss = self.training_config["loss"]
+        if loss == "L1":
+            self.loss = nn.L1Loss()
+        elif loss == "L2":
+            self.loss = nn.MSELoss()
```
jarlsondre, line 410: Double check that capital letters are used for the loss everywhere, i.e. "L1" instead of "l1". I remember there being something in EURAC or something where these things were different.

src/itwinai/torch/distributed.py
```diff
@@ -1010,3 +1012,177 @@ def gather(self, tensor: torch.Tensor, dst_rank: int = 0):
...
+        pass
+
+    def gather_obj(self, obj: Any, dst_rank: int = 0) -> List[Any]:
+        pass
+
+    def gather(self, tensor: torch.Tensor, dst_rank: int = 0) -> List | None:
+        pass
+
+
+class RayDeepSpeedStrategy(RayDistributedStrategy):
```
jarlsondre, line 1124: Would it be possible for this strategy to inherit "DeepSpeedStrategy" (or whatever it is called) instead? Would that make any sense at all?

use-cases/virgo/trainer.py
```diff
@@ -373,7 +370,279 @@ def train(self):
...
+        """Override the create_dataloaders function to use the custom_collate function."""
+        # This is the case if a small dataset is used in-memory
+        # - we can use the default collate_fn function
+        if isinstance(train_dataset, TensorDataset):
+            return super().create_dataloaders(
+                train_dataset=train_dataset,
+                validation_dataset=validation_dataset,
+                test_dataset=test_dataset,
+            )
+        else:
```
jarlsondre, line 450: You can omit the else, since you return earlier.

use-cases/virgo/trainer.py
```diff
@@ -373,7 +370,279 @@ def train(self):
...
+                    batch_size=self.training_config["batch_size"],
+                    num_workers=self.training_config["num_workers_dataloader"],
+                    pin_memory=self.training_config["pin_gpu_memory"],
+                    # generator=self.torch_rng,
+                    shuffle=self.training_config["shuffle_test"],
+                    collate_fn=self.custom_collate,
+                )
+
+    def custom_collate(self, batch):
+        """
```
jarlsondre, line 484: remove newline

use-cases/virgo/trainer.py
```diff
@@ -373,7 +370,279 @@ def train(self):
...
+        Custom collate function to concatenate input tensors along their first dimension.
+        """
+        # Some batches contain None values, if any files from the dataset did not match the criteria
+        # (i.e. three auxilliary channels)
+        batch = [x for x in batch if x is not None]
+
+        return torch.cat(batch)
+
+    def train(self, config, data):
```
jarlsondre, line 493: This function has a lot of things that could be cleaned up, but I understand that it isn't your work so I'll let you decide if you want me to actually point everything out :D

src/itwinai/torch/distributed.py
annaelisalappe: Yup :)

use-cases/virgo/trainer.py
```diff
@@ -373,7 +370,279 @@ def train(self):
...
+            self.model = Decoder_2d_deep(3)
+            init_weights(self.model, "normal", scaling=0.02)
+        elif generator == "resnet":
+            self.model = GeneratorResNet(3, 12, 1)
+            init_weights(self.model, "normal", scaling=0.01)
+        elif generator == "unet":
+            self.model = UNet(input_channels=3, output_channels=1, norm=False)
+            init_weights(self.model, "normal", scaling=0.02)
+        else:
+            raise ValueError("Unrecognized generator type! Got", generator)
```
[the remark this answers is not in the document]
annaelisalappe: Changed this in both trainers

use-cases/virgo/trainer.py
annaelisalappe: Yes this was always inconsistent in Virgo, I have changed everything to lowercase

src/itwinai/torch/distributed.py
annaelisalappe: Yeah I thought about this, it would definitely save some lines of code as many functions are exactly the same, e.g. `distributed` and `global/local_world_size`, etc. On the other hand we would have to then move the ray initialisation logic into the RayDeepSpeedStrategy.. Also, at least imo, it's nicer to have all ray strategies uniformly inheriting from the same strategy, but that is more a design choice, both would work. I would however, if we changed this, remove the RayDistributedStrategy altogether and have all ray strategies inherit from their pure torch counterpart. There is also some frameworks that ray supports and itwinai currently does not (huggingface transformers/ accelerate for example), that I thought we might want to implement in the future, in which case they would not have an itwinai torch strategy counterpart

use-cases/virgo/trainer.py
annaelisalappe: Yeah I think I would leave the refactoring of the code that the virgo team wrote for another time, I would rather first make sure that the ray stuff is merged 😌

README.md
matbun: you're right

annaelisalappe: So do you think I should change this here or should I create an issue to change this across the codebase?

src/itwinai/torch/distributed.py
matbun: I would relax a bit the structure by transforming `_initialize_ray` method in a function and removing the `RayDistributedStrategy`.

I don't think that there is much in common between DDP and RayDDP, so RayDDP could inherit directly from TorchDistributedStrategy. 

Regarding deepspeed, I would try to reuse as much as possible from the DeepSpeedStrategy class already there, thus having RayDeepSpeedStrategy to inherit from DeepSpeedStrategy.

I agree that it would be nice to have an overarching Ray class, and we could actually have it (RayDeepSpeedStrategy  could inherit both from RayDistributedStrategy and DeepSpeedStrategy). But maybe it's complicating things a bit?

What do you think?

matbun:
> * Adding further type checks to the different configs, as well as changing the current dictionary setup to something using a class that implements the `__get_item()__` method

See how we implemented the training config with pydantic. It could be an interesting input

README.md
r-sarma: This will not necessarily remain at Python 3.11 in the next stage. We will soon have Stage 2025, when the version will potentially change. And the new stage will probably provide the latest Python version. For HPC centers, these scripts have to be anyway updated with time.

src/itwinai/torch/distributed.py
```diff
@@ -3,10 +3,13 @@
 import os
 from typing import Any, Callable, Iterable, List, Literal, Optional, Tuple, Union
 
+import ray
+import ray.train
```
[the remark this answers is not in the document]
annaelisalappe: I added ray[default, train, tune] to the pyproject.toml under _dependencies_

src/itwinai/torch/distributed.py
```diff
@@ -1010,3 +1007,180 @@ def gather(self, tensor: torch.Tensor, dst_rank: int = 0):
...
+    def local_rank(self) -> int:
+        return ray.train.get_context().get_local_rank()
+
+    @check_initialized
+    def distributed(
+        self,
+        model: nn.Module,
+        optimizer: Optimizer,
+        lr_scheduler: Optional[LRScheduler] = None,
+    ) -> Tuple[Module, Optimizer, LRScheduler | None]:
```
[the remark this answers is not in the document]
annaelisalappe: Good point, changed this :)

src/itwinai/torch/distributed.py
```diff
@@ -1010,3 +1007,180 @@ def gather(self, tensor: torch.Tensor, dst_rank: int = 0):
...
         """
         return [tensor]
+
+
+class RayDistributedStrategy(TorchDistributedStrategy):
+    def _initialize_ray(self) -> None:
+        if not ray.is_initialized():
+            try:
+                ip_head = os.environ.get("ip_head")
+                head_node_ip = os.environ.get("head_node_ip")
```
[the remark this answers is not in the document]
annaelisalappe: Yes these are set in the slurm script to launch the ray cluster, I've changed them to uppercase

src/itwinai/torch/distributed.py
```diff
@@ -1010,3 +1007,180 @@ def gather(self, tensor: torch.Tensor, dst_rank: int = 0):
...
+        pass
+
+    def allgather_obj(self, obj: Any) -> List[Any]:
+        pass
+
+    def gather_obj(self, obj: Any, dst_rank: int = 0) -> List[Any]:
+        pass
+
+    def gather(self, tensor: torch.Tensor, dst_rank: int = 0) -> List | None:
+        pass
```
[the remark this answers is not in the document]
annaelisalappe: This is one of the things that I wanted to look at in a future PR, I had really not thought about these methods and because so far in the virgo integration they haven't come up I thought it would be okay to leave them like this for now?

src/itwinai/torch/distributed.py
annaelisalappe: I definitely see your point - we could move the inheritance, but I would keep it uniform and have the ray strategies inherit from their torch counterpart. The RayDDPStrategy would be overwriting most of the functions, but (to me, at least) it would be better style to have it uniform, otherwise I think the inheritance would be confusing, especially when we add more strategies. 

And then for the second point, I think it's hard to tell now if having a separate RayDistributedStrategy makes sense, because 
a. there is at least two more strategies that I think would be cool to add (horovod and huggingface accelerate), and I don't know yet which methods they need and if anything else could be consolidated in a shared parent class
b. I have not had a chance to look at some functions, such as `gather` and `clean_up`, so same for these

So I would suggest to leave the RayDistributedStrategy for now, and if you think it would be better, I can add multi-inheritance to make the RayDDPStrategy inherit from the DDPStrategy and the RayDeepSpeedStrategy to inherit from the DeepSpeedStrategy.

src/itwinai/torch/distributed.py
```diff
@@ -1010,3 +1007,180 @@ def gather(self, tensor: torch.Tensor, dst_rank: int = 0):
...
+                        "Ray initialization requires 'ip_head' and 'head_node_ip' to be set."
+                    )
+
+            except Exception as e:
+                raise RuntimeError(f"Error initializing Ray: {str(e)}")
+
+            ray.init(address="auto")
+
+
+class RayDDPStrategy(RayDistributedStrategy):
```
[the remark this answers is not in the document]
annaelisalappe: Added

src/itwinai/torch/distributed.py
```diff
@@ -1010,3 +1007,180 @@ def gather(self, tensor: torch.Tensor, dst_rank: int = 0):
...
+        pass
+
+    def gather_obj(self, obj: Any, dst_rank: int = 0) -> List[Any]:
+        pass
+
+    def gather(self, tensor: torch.Tensor, dst_rank: int = 0) -> List | None:
+        pass
+
+
+class RayDeepSpeedStrategy(RayDistributedStrategy):
```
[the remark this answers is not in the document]
annaelisalappe: Added

src/itwinai/torch/trainer.py
```diff
@@ -1212,42 +1180,420 @@ def distributed(func):
...
+
+    @property
+    def device(self) -> str:
+        """Get the current device from distributed strategy.
+        Returns:
+            str: Device string (e.g., "cuda:0").
+        """
+        return self.strategy.device()
+
+    def create_dataloaders(
```
matbun, line 1326: Who is calling this method? Is this used?

use-cases/virgo/trainer.py
```diff
@@ -373,7 +367,278 @@ def train(self):
...
+            )
+        else:
+            distribute_kwargs = {}
+
+        # Distributed model, optimizer, and scheduler
+        self.model, self.optimizer, _ = self.strategy.distributed(
+            self.model, self.optimizer, **distribute_kwargs
+        )
+
+    def create_dataloaders(
```
matbun, line 444: could this method be moved in the parent RayTorchTrainer?

src/itwinai/torch/distributed.py
annaelisalappe: We can still refine it in the future, if we add more strategies 😌

pyproject.toml
```diff
@@ -37,6 +37,7 @@ dependencies = [
     "prov4ml@git+https://github.com/matbun/ProvML@new-main",
     "pandas",
     "seaborn",
+    "ray[default, train, tune]",
```
matbun, line 40: I think the spaces may cause some troubles

src/itwinai/torch/trainer.py
```diff
@@ -1220,3 +1231,382 @@ def dist_train(
...
+            collate_fn (Callable[[List], Any], optional):
+                Function to collate data samples into batches. Defaults to None.
+        """
+        self.train_dataloader = self.strategy.create_dataloader(
+            dataset=train_dataset,
+            batch_size=batch_size,
+            shuffle=shuffle_train,
+            sampler=sampler,
+            collate_fn=collate_fn,
+        )
```
matbun, line 1362: why the torch_rng is not passed to the dataloader? 
Why does this diverge from the implementation in the TorchTrainer (see below)?

```python
self.train_dataloader = self.strategy.create_dataloader(
            dataset=train_dataset,
            batch_size=self.config.batch_size,
            num_workers=self.config.num_workers_dataloader,
            pin_memory=self.config.pin_gpu_memory,
            generator=self.torch_rng,
            shuffle=self.config.shuffle_train,
        )
```

Same for the code blocks below

use-cases/virgo/trainer.py
```diff
@@ -385,7 +379,278 @@ def train(self):
...
+            )
+
+        # If we are using a custom dataset for the large dataset,
+        # we need to overwrite the collate_fn function
+        self.train_dataloader = self.strategy.create_dataloader(
+            dataset=train_dataset,
+            batch_size=self.training_config["batch_size"],
+            num_workers=self.training_config["num_workers_dataloader"],
+            pin_memory=self.training_config["pin_gpu_memory"],
+            # generator=self.torch_rng,
```
matbun, line 467: it would be nice to pass the PRNG for reproducibility

use-cases/virgo/trainer.py
```diff
@@ -385,7 +379,278 @@ def train(self):
...
+                shuffle=self.training_config["shuffle_validation"],
+                collate_fn=self.custom_collate,
+            )
+        if test_dataset is not None:
+            self.test_dataloader = self.strategy.create_dataloader(
+                dataset=test_dataset,
+                batch_size=self.training_config["batch_size"],
+                num_workers=self.training_config["num_workers_dataloader"],
+                pin_memory=self.training_config["pin_gpu_memory"],
+                # generator=self.torch_rng,
```
matbun, line 487: it would be nice to pass the PRNG for reproducibility

matbun (changes requested):

src/itwinai/torch/distributed.py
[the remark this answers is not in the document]
matbun: I think this is a bit too critical part to skip before merging, because developing it may change a lot the structure of the Ray strategies

src/itwinai/torch/distributed.py
```diff
@@ -56,6 +58,28 @@ def wrapper(self: "TorchDistributedStrategy", *args, **kwargs):
...
+    the Ray backend if it is not already initialized.
+
+        Raises:
+            EnvironmentError: If required environment variables `IP_HEAD` or `HEAD_NODE_IP`
+                are not set. These should be set from the slurm script where the ray cluster
+                is launched.
+    """
+    if not ray.is_initialized():
+        IP_HEAD = os.environ.get("IP_HEAD")
+        HEAD_NODE_IP = os.environ.get("HEAD_NODE_IP")
```
matbun, line 72: what's the difference in Ray between these env vars? The names are a bit confusing because it seems that they are both the IP of the head node

src/itwinai/torch/raytune.py
```diff
@@ -0,0 +1,123 @@
...
+            seed = 1234
+        else:
+            seed = None
+        return TuneBOHB(
+            metric=metric,
+            mode=mode,
+            seed=seed,
+        )
+
+    # requires pip install bayesian-optimization
```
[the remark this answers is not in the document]
annaelisalappe: So adding all into one dedicated group in the pyproject.toml? So far I have been installing the one I need directly in the slurm script, because I think that adding a dedicated dependency group for each of them is too much and you might not want to install all of them just to use one, if that makes sense. But I could definitely add them, to be fair they should all be pretty fast to install...

src/itwinai/torch/trainer.py
```diff
@@ -1212,42 +1180,420 @@ def distributed(func):
...
+        "loss": "cross_entropy",
+        "optim_momentum": 0.9,
+        "optim_weight_decay": 0,
+        "num_workers_dataloader": 4,
+        "random_seed": 21,
+    },
+}
+
+
+class RayTorchTrainer(Trainer):
```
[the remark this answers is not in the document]
annaelisalappe: Added

src/itwinai/torch/raytune.py
```diff
@@ -0,0 +1,123 @@
...
+    HyperBandScheduler,
+    PopulationBasedTraining,
+)
+from ray.tune.schedulers.pb2 import PB2  # Population Based Bandits
+from ray.tune.search.bayesopt import BayesOptSearch
+from ray.tune.search.bohb import TuneBOHB
+from ray.tune.search.hyperopt import HyperOptSearch
+
+
+def get_raytune_search_alg(tune_config, seeds=False):
```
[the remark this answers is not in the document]
annaelisalappe: Added

use-cases/virgo/trainer.py
annaelisalappe: This is the same as in the other Virgo Trainer, because we use a custom collate function this needs to be redefined here

src/itwinai/torch/trainer.py
annaelisalappe: Yes this is the same as for the TorchTrainer, it might be used if a use case is fine using the default implementation. If we change the inheritance it could of course be removed as it is the same as the `create_dataloaders` function of the TorchTrainer

src/itwinai/torch/trainer.py
```diff
@@ -1212,42 +1180,420 @@ def distributed(func):
...
+                sampler=sampler,
+                collate_fn=collate_fn,
+            )
+        else:
+            self.test_dataloader = None
+
+    @abstractmethod
+    def train(config, data=None):
+        pass
+
```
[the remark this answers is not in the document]
annaelisalappe: Okay this is a good place to discuss moving the inheritance from the `Trainer` to the `TorchTrainer`. I definitely agree that the duplication of many methods, such as `create_dataloaders` or `set_epoch`  are not great. On the other hand, most methods in the `TorchTrainer` do not quite work with the `RayTorchTrainer`, mostly for the RayDDPStrategy, which is the default strategy provided by ray for training. For example, the moving of data to devices is done automatically when creating the DataLoader, and you can't call `enumerate(self.dataloader)`, that will throw an error. So most of the functions that are specific to training, such as `train`, `train_epoch`, etc, would have to be overwritten. For now I would prefer to leave them empty, so I could do something like the above 
```
@abstractmethod
    def train(config, data=None):
        pass
```
for all of the functions that would not work as-is with the `RayTorchTrainer`. 

Do you have a preference, as in having the RayTorchTrainer inherit from the TorchTrainer with this in mind, or leaving it the way it is?

src/itwinai/torch/trainer.py
```diff
@@ -1212,42 +1180,420 @@ def distributed(func):
...
+        result_grid = tuner.fit()
+
+        return train_dataset, validation_dataset, test_dataset, result_grid
+
+    def set_epoch(self, epoch: int) -> None:
+        self.train_dataloader.sampler.set_epoch(epoch)
+        if self.validation_dataloader is not None:
+            self.validation_dataloader.sampler.set_epoch(epoch)
+        if self.test_dataloader is not None:
+            self.test_dataloader.sampler.set_epoch(epoch)
```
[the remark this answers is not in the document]
annaelisalappe: I will close these to move the conversation to a central place above :)

src/itwinai/torch/trainer.py
annaelisalappe: I kept this more minimal for now because I had not made a final design choice about the config, and the dictionary implementation so far does not support default values. As this will change anyway, is it okay if I just make a note of this to change once I figure out the config situation?

src/itwinai/torch/distributed.py
```diff
@@ -1010,3 +1007,180 @@ def gather(self, tensor: torch.Tensor, dst_rank: int = 0):
...
+        pass
+
+    def allgather_obj(self, obj: Any) -> List[Any]:
+        pass
+
+    def gather_obj(self, obj: Any, dst_rank: int = 0) -> List[Any]:
+        pass
+
+    def gather(self, tensor: torch.Tensor, dst_rank: int = 0) -> List | None:
+        pass
```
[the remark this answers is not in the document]
annaelisalappe: Closing this to move the conversation to a central place

src/itwinai/torch/distributed.py
annaelisalappe: Good point, I have renamed IP_HEAD to HEAD_NODE_PORT to make the difference clearer.

src/itwinai/torch/distributed.py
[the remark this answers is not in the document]
annaelisalappe: Okay I changed the RayDDPStrategy that it now inherits from the TorchDDPStrategy - I think the way the strategies are organised now is the best way to do it, and it turns out, `gather`, `allgather_obj` and `gather_obj` could just be reused from their parent strategies (I have tested them in both DDP and DeepSpeed and both work).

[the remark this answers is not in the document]
annaelisalappe: I had looked at the ray collective library but I believe that this is geared more towards usage with ray core and not needed when we are using the torch distributed backend (which is great, because it would complicate our code so much 🙌🏼)

.github/workflows/lint.yml
```diff
@@ -40,6 +40,7 @@ jobs:
 
           VALIDATE_PYTHON: true
           VALIDATE_PYTHON_BLACK: false
+          VALIDATE_ISORT: false
```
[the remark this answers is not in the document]
matbun: ok, thanks

src/itwinai/torch/distributed.py
```diff
@@ -1015,3 +1039,93 @@ def gather(self, tensor: torch.Tensor, dst_rank: int = 0):
...
+        timeout: float = 0,
+        worker_init_fn: Optional[_worker_init_fn_t] = None,
+        multiprocessing_context=None,
+        generator=None,
+        *,
+        prefetch_factor: Optional[int] = None,
+        persistent_workers: bool = False,
+        pin_memory_device: str = "",
+    ):
+        dataloader = super().create_dataloader(
```
matbun, line 1101: be careful that the `super().create_dataloader(...)` method will create a data loader *already* with a distributed sampler, which may not be what you want to do when using Ray (e.g., because it may not be needed). If you don't need this, replace `super().create_dataloader(...)` simply with `Dataloader`

annaelisalappe: Good catch, thanks!

use-cases/virgo/trainer.py
annaelisalappe: Okay added!

use-cases/virgo/trainer.py
annaelisalappe: Okay added!

src/itwinai/torch/trainer.py
annaelisalappe: Okay I decided to make it more in line with the TorchTrainer implementation anyway, even if we change it later

src/itwinai/torch/distributed.py
```diff
@@ -56,6 +59,28 @@ def wrapper(self: "TorchDistributedStrategy", *args, **kwargs):
...
+def _initialize_ray() -> None:
+    """This method is used by the RayDDPStrategy and RayDeepSpeedStrategy to initialize
+    the Ray backend if it is not already initialized.
+
+        Raises:
+            EnvironmentError: If required environment variables `HEAD_NODE_PORT` or
+                `HEAD_NODE_IP` are not set.
+                These should be set from the slurm script where the ray cluster is launched.
+    """
+    if not ray.is_initialized():
```
jarlsondre, line 71: maybe negate this and return early, avoiding the indent?

src/itwinai/torch/trainer.py
```diff
@@ -1220,3 +1231,390 @@ def dist_train(
                 dist.destroy_process_group()
 
     return dist_train
+
+
+DEFAULT_RAY_CONFIG = {
```
jarlsondre, line 1236: In general, I dislike the idea of having constants like this in the middle of a file. This clutters the file a bit and makes it harder to maintain in the future. I think I would prefer either 

1) putting it in a different file (`constants.py` or something) or 

2) at least putting it at the top of this file.

src/itwinai/torch/trainer.py
```diff
@@ -1220,3 +1231,390 @@ def dist_train(
...
+            self.tune_config = tune.TuneConfig(
+                num_samples=num_samples,
+                metric=metric if not search_alg and not scheduler else None,
+                mode=mode if not search_alg and not scheduler else None,
+                search_alg=search_alg,
+                scheduler=scheduler,
+            )
+        else:
+            print("INFO: No Tune Config configured.")
+            self.tune_config = None
```
jarlsondre, line 1462: If you do the `else` first and then return, you can avoid a lot of indentation here.

src/itwinai/torch/trainer.py
```diff
@@ -1220,3 +1231,390 @@ def dist_train(
...
+            )
+            scheduler = (
+                get_raytune_schedule(tune_config) if "scheduler" in tune_config else None
+            )
+
+            # Only set metric and mode if search_alg and scheduler aren't defined
+            self.tune_config = tune.TuneConfig(
+                num_samples=num_samples,
+                metric=metric if not search_alg and not scheduler else None,
+                mode=mode if not search_alg and not scheduler else None,
```
jarlsondre, line 1456: Personally I think these are a little too convoluted to do inline. I would make variables before `self.tune_config = ...`

src/itwinai/torch/trainer.py
```diff
@@ -1220,3 +1231,390 @@ def dist_train(
...
+
+    def _set_run_config(self):
+        run_config = self.config.get("run_config")
+        if run_config:
+            storage_path = Path(run_config.get("storage_path"))
+            if storage_path:
+                self.run_config = ray.train.RunConfig(storage_path=Path.absolute(storage_path))
+        else:
+            print("INFO: No RunConfig provided. Assuming local or single-node execution.")
+            self.run_config = None
```
jarlsondre, line 1480: In this case it seems to me that you are checking if `run_config` is in the config and that you are only setting `self.run_config` if `storage_path` is a field of it. I had some trouble understanding what was going on here for a second and thus I think I would prefer something like: 

```python
if not "run_config" in self.config: 
    self.run_config = None
    print(...)
    return

run_config = self.config["run_config"]
if not "storage_path" in run_config: 
   return

self.run_config = ray.train...
```

src/itwinai/torch/trainer.py
```diff
@@ -1220,3 +1231,390 @@ def dist_train(
...
+    def _set_train_loop_config(self):
+        train_loop_config = self.config.get("train_loop_config")
+        if train_loop_config:
+            self.train_loop_config = self._set_searchspace(train_loop_config)
+        else:
+            print(
+                "INFO: No training_loop_config detected. "
+                "No parameters are being tuned or passed to the training function."
+            )
+            self.train_loop_config = {}
```
jarlsondre, line 1491: Same as above, I would use `in` to streamline this function a bit.

src/itwinai/torch/trainer.py
```diff
@@ -1220,3 +1231,390 @@ def dist_train(
...
+            if checkpointing_data and should_checkpoint:
+                torch.save(checkpointing_data, os.path.join(temp_checkpoint_dir, str(epoch)))
+                checkpoint = train.Checkpoint.from_directory(temp_checkpoint_dir)
+
+        train.report(tuning_metrics, checkpoint=checkpoint)
+
+    def initialize_logger(self, hyperparams: Optional[Dict], rank):
+        print(f"Logger initializing with rank {rank}")
+
+        if self.logger:
```
jarlsondre, line 1573: You can negate this. Also, are you sure you want to print the above statement even if `self.logger` is None?

src/itwinai/torch/tuning.py
```diff
@@ -0,0 +1,158 @@
...
+        An instance of the chosen Ray Tune search algorithm or None if no search algorithm is
+            used or if the search algorithm does not match any of the supported options.
+
+    Notes:
+        - `TuneBOHB` is automatically chosen for BOHB scheduling.
+    """
+    if "scheduler" in tune_config:
+        scheduler = tune_config["scheduler"]["name"]
+    else:
+        scheduler = None
```
jarlsondre, line 34: Ironically, here you can use `.get()` to avoid some lines

src/itwinai/torch/tuning.py
```diff
@@ -0,0 +1,158 @@
...
+            print(
+                "INFO: Using schedule '{}' \
+                    is not compatible with Ray Tune search algorithms.".format(scheduler)
+            )
+            print(
+                "INFO: Using the Ray Tune '{}' scheduler without search algorithm".format(
+                    scheduler
+                )
+            )
+        return None
```
jarlsondre, line 50: Here you can negate and return early to avoid indentation

use-cases/virgo/trainer.py
```diff
@@ -385,7 +379,233 @@ def train(self):
...
+        config: Dict,
+        strategy: Optional[Literal["ddp", "deepspeed"]] = "ddp",
+        name: Optional[str] = None,
+        logger: Optional[Logger] = None,
+    ) -> None:
+        super().__init__(config=config, strategy=strategy, name=name, logger=logger)
+
+    def create_model_loss_optimizer(self) -> None:
+        # Select generator
+        generator = self.training_config["generator"]
```
jarlsondre, line 398: Just want to double check since I've seen you check these things before: Can you be sure that `self.training_config` contains "generator" here? 😁

src/itwinai/torch/trainer.py
annaelisalappe: Okay I'll move it to the top, but keep in mind this is a temporary fix before I figure out how to handle the config better 😌

src/itwinai/torch/trainer.py
annaelisalappe: Changed this

use-cases/virgo/trainer.py
```diff
@@ -59,7 +59,7 @@ def __init__(
     ) -> None:
         super().__init__(
             epochs=num_epochs,
-            config=config,
+            config={},
```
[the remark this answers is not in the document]
annaelisalappe: Okay, but does this matter if it's an empty dictionary? I will change it regardless though, I actually didn't know this :)

use-cases/virgo/trainer.py
annaelisalappe: Hmm well no not exactly, so far this has to be defined in the config.yaml file, otherwise there will be an error. This is one of the reasons I will change the structure with the dictionary object in a future PR, for type-checking and checking that all fields are set.

src/itwinai/torch/tuning.py
annaelisalappe: Okay I've changed this, I find it a bit more convoluted though tbh, what do you think?
`scheduler = tune_config.get("scheduler", {}).get("name")`

src/itwinai/torch/raytune.py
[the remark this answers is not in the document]
annaelisalappe: Okay I've added them in a dedicated 'hpo' group in the `pyproject.toml`, let me know if you would like me to change it

pyproject.toml
```diff
@@ -65,6 +66,14 @@ docs = [
     "IPython",
     "tensorflow==2.16.*",
 ]
+hpo = [
+    "bayesian-optimization==2.0.0",
+    "hyperopt==0.2.7",
+    "ConfigSpace==1.2.1",
+    "hpbandster==0.7.4",
+    "scipy==1.14.1",
+    "GPy==1.13.2"
```
matbun, line 75: I would avoid fixing the patch version when using `==`. Example: `hyperopt==0.2.*`

src/itwinai/torch/trainer.py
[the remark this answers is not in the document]
matbun: I see. Forcing some methods to be abstract just to avoid inheriting them is not a good idea, because the class cannot be instantiated without overriding *all* the abstract methods, even if unused. Perhaps a better idea would be moving the methods which are in common between RayTorchTrainer and TorchTrainer in a common parent class, e.g. BaseTorchTrainer, form which both trainers inherit

src/itwinai/torch/trainer.py
```diff
@@ -1212,42 +1180,420 @@ def distributed(func):
...
+                item=item,
+                identifier=identifier,
+                kind=kind,
+                step=step,
+                batch_idx=batch_idx,
+                **kwargs,
+            )
+        else:
+            print(
+                "INFO: The log method was called, but no logger was configured for this \
```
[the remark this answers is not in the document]
matbun: seems still unsolved

use-cases/virgo/slurm_ray.sh
```diff
@@ -47,9 +48,13 @@ nodes_array=($nodes)
 head_node=${nodes_array[0]}
 port=7639       # This port will be used by Ray to communicate with worker nodes.
 
-# This is so that the ray.init() command called from the hpo.py script knows which ports to connect to
-export ip_head="$head_node"i:"$port"
-export head_node_ip="$head_node"i
+# This is so that the ray.init() command called from the hpo.py script knows
+# which ports to connect to
+export HEAD_NODE_PORT="$head_node"i:"$port"
```
matbun, line 53: wait, here `HEAD_NODE_PORT` seems a concat of both IP and PORT.

I like the idea of having a varible called `HEAD_NODE_PORT` (see also two lines below for another example), but it must be just the port. In the python script you can concat HEAD_NODE_PORT and HEAD_NODE_IP to correctly pass the info to Ray

src/itwinai/torch/distributed.py
```diff
@@ -56,6 +59,30 @@ def wrapper(self: "TorchDistributedStrategy", *args, **kwargs):
...
+        Raises:
+            EnvironmentError: If required environment variables `HEAD_NODE_PORT` or
+                `HEAD_NODE_IP` are not set.
+                These should be set from the slurm script where the ray cluster is launched.
+    """
+    if ray.is_initialized():
+        return
+
+    HEAD_NODE_PORT = os.environ.get("HEAD_NODE_PORT")
+    HEAD_NODE_IP = os.environ.get("HEAD_NODE_IP")
```
matbun, line 75: where are they used?

use-cases/virgo/trainer.py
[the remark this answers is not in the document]
jarlsondre: Depending on how often it is called, it might not matter at all, but the safest action is to avoid it :)

src/itwinai/torch/tuning.py
jarlsondre: I agree with you, it was better the way it was before. I missed the fact that you have the double-dictionary structure. That does beg the question though: Why is it called "scheduler" when you have to extract the "name" from the scheduler? Maybe the variable should be called "scheduler_name" or something instead? Not super important haha

use-cases/virgo/slurm_ray.sh
annaelisalappe: Resolving because of below comment

pyproject.toml
annaelisalappe: Okay I'll change it

src/itwinai/torch/distributed.py
annaelisalappe: Okay the official ray script exports 
```
port=6379
ip_head=$head_node_ip:$port
export ip_head
echo "IP Head: $ip_head"
```
but not 
`export head_node_ip="$head_node"I`
that is only in Eric's script. To be honest I don't know why he has it

matbun: after a quick look online it seems to me that these env vars are useless in the python code. I would simply remove them and see if the code runs :)

pyproject.toml
matbun: if you do `>=` it's fine, but with `==` you are not allowing new patches to the same version. You can fix this in cases you see `==` and the patch version

src/itwinai/torch/distributed.py
annaelisalappe: I'll push a new version where I have removed `export head_node_ip="$head_node"I` and renamed the variables to their original version. I agree that maybe the names are not the most explicit, but they are used consistently this way in all examples and in the documentation, so I feel like ensuring consistency here is more important, if we rename the variables it confuses things too much in my opinion 😌 But good that you pointed it out! I had just copied Eric's code but it could be simplified a bit apparently

matbun: yes, I would remove all exports whatsoever. They don't seem to be needed anywhere

README.md
matbun: ok, so for now let's leave it as is

src/itwinai/torch/tuning.py
annaelisalappe: Hmm I see your point, but there is a few more variables that need to be defined under "scheduler", it could of course be named something else, but the value of "scheduler" needs to be a dictionary. I will resolve this because we talked about it in-person :)

src/itwinai/torch/distributed.py
annaelisalappe: Okay, this is done in the new version :)

pyproject.toml
annaelisalappe: Added!

matbun (approved):

PR #261, opened 2024-11-27
Add Path in loggers and clearer error handling

### LINKED ISSUE #259: Errors in pipeline are hard to read
When an error happens while using the pipeline, the error stack is not very user friendly

matbun:
# Summary

Improve the readability of errors when using the itwinai config parser, for instance when executing a pipeline from configuration file.

---

**Related issue :**

Closes #259

### DIFF
--- a/src/itwinai/loggers.py
+++ b/src/itwinai/loggers.py
@@ -707,7 +707,10 @@ class WandBLogger(Logger):
         if not self.should_log():
             return
 
-        (self.savedir / "wandb").mkdir(exist_ok=True)
+        (self.savedir / "wandb").mkdir(
+            exist_ok=True,
+            parents=True,
+        )
         self.active_run = wandb.init(dir=self.savedir.resolve(), project=self.project_name)
 
     def destroy_logger_context(self):
@@ -803,7 +806,7 @@ class TensorBoardLogger(Logger):
             import tensorflow as tf
 
             self.tf = tf
-            self.writer = tf.summary.create_file_writer(savedir)
+            self.writer = tf.summary.create_file_writer(savedir.resolve().as_posix())
         elif framework.lower() == "pytorch":
             from torch.utils.tensorboard import SummaryWriter
 
--- a/src/itwinai/loggers.py
+++ b/src/itwinai/loggers.py
@@ -458,24 +458,6 @@ class ConsoleLogger(Logger):
             print(f"ConsoleLogger: {identifier} = {item}")
 
 
-def root():
-    raise ValueError("ROOT")
-
-
-def middle():
-    try:
-        root()
-    except Exception:
-        raise ValueError("MIDDLE")
-
-
-def outer():
-    try:
-        middle()
-    except Exception:
-        raise ValueError("OUTER")
-
-
 class MLFlowLogger(Logger):
     """Abstraction around MLFlow logger.
 
--- a/src/itwinai/loggers.py
+++ b/src/itwinai/loggers.py
@@ -113,7 +113,7 @@ class Logger(LogMixin):
     """Base class for logger
 
     Args:
-        savedir (Path, optional): filesystem location where logs are stored.
+        savedir (Union[Path, str], optional): filesystem location where logs are stored.
             Defaults to 'mllogs'.
         log_freq (Union[int, Literal['epoch', 'batch']], optional):
             how often should the logger fulfill calls to the `log()`
@@ -136,7 +136,7 @@ class Logger(LogMixin):
     """
 
     #: Location on filesystem where to store data.
-    savedir: Optional[Path] = None
+    savedir: Path
     #: Supported logging 'kind's.
     supported_kinds: Tuple[str]
     #: Current worker global rank
@@ -146,13 +146,13 @@ class Logger(LogMixin):
 
     def __init__(
         self,
-        savedir: Path = "mllogs",
+        savedir: Path | str = "mllogs",
         log_freq: Union[int, Literal["epoch", "batch"]] = "epoch",
         log_on_workers: Union[int, List[int]] = 0,
         experiment_id: Optional[str] = None,
         run_id: Optional[Union[int, str]] = None,
     ) -> None:
-        self.savedir = savedir
+        self.savedir = Path(savedir)
         self.log_freq = log_freq
         self.log_on_workers = log_on_workers
         self._experiment_id = experiment_id
@@ -240,7 +240,7 @@ class Logger(LogMixin):
         Returns:
             str: local path of the serialized object to be logged.
         """
-        itm_path = Path(self.savedir) / str(identifier)
+        itm_path = self.savedir / str(identifier)
         with open(itm_path, "wb") as itm_file:
             pickle.dump(obj, itm_file)
 
@@ -301,7 +301,7 @@ class _EmptyLogger(Logger):
 
     def __init__(
         self,
-        savedir: Path = "mllogs",
+        savedir: Path | str = "mllogs",
         log_freq: int | Literal["epoch"] | Literal["batch"] = "epoch",
         log_on_workers: int | List[int] = 0,
     ) -> None:
@@ -332,7 +332,7 @@ class ConsoleLogger(Logger):
     """Simplified logger.
 
     Args:
-        savedir (Path, optional): where to store artifacts.
+        savedir (Union[Path, str], optional): where to store artifacts.
             Defaults to 'mllogs'.
         log_freq (Union[int, Literal['epoch', 'batch']], optional):
             determines whether the logger should fulfill or ignore
@@ -349,7 +349,7 @@ class ConsoleLogger(Logger):
 
     def __init__(
         self,
-        savedir: Path = "mllogs",
+        savedir: Path | str = "mllogs",
         log_freq: Union[int, Literal["epoch", "batch"]] = "epoch",
         log_on_workers: Union[int, List[int]] = 0,
     ) -> None:
@@ -462,7 +462,7 @@ class MLFlowLogger(Logger):
     """Abstraction around MLFlow logger.
 
     Args:
-        savedir (Path, optional): path on local filesystem where logs are
+        savedir (Union[Path, str], optional): path on local filesystem where logs are
             stored. Defaults to 'mllogs'.
         experiment_name (str, optional): experiment name. Defaults to
             ``itwinai.loggers.BASE_EXP_NAME``.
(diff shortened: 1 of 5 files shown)


src/itwinai/loggers.py
```diff
@@ -136,7 +136,7 @@ class Logger(LogMixin):
     """
 
     #: Location on filesystem where to store data.
-    savedir: str = None
+    savedir: Optional[Path] = None
```
annaelisalappe, line 139: Maybe this would be a good place to start using `Path | None`?

src/itwinai/loggers.py
```diff
@@ -240,7 +240,7 @@ def serialize(self, obj: Any, identifier: str) -> str:
         Returns:
             str: local path of the serialized object to be logged.
         """
-        itm_path = os.path.join(self.savedir, identifier)
+        itm_path = Path(self.savedir) / str(identifier)
```
annaelisalappe, line 243: If `self.savedir` is already a Path, shouldn't `self.savedir / str(identifier)` be sufficient here?

src/itwinai/loggers.py
```diff
@@ -349,11 +349,11 @@ class ConsoleLogger(Logger):
 
     def __init__(
         self,
-        savedir: str = "mllogs",
+        savedir: Path = "mllogs",
         log_freq: Union[int, Literal["epoch", "batch"]] = "epoch",
         log_on_workers: Union[int, List[int]] = 0,
     ) -> None:
-        savedir = savedir = Path(savedir) / "simple-logger"
+        savedir = Path(savedir) / "simple-logger"
```
annaelisalappe, line 356: Same here, is `savedir / "simple_logger"` sufficient?

src/itwinai/loggers.py
```diff
@@ -501,15 +501,15 @@ class MLFlowLogger(Logger):
...
+        savedir: Path = "mllogs",
         experiment_name: str = BASE_EXP_NAME,
         tracking_uri: Optional[str] = None,
         run_description: Optional[str] = None,
         run_name: Optional[str] = None,
         log_freq: Union[int, Literal["epoch", "batch"]] = "epoch",
         log_on_workers: Union[int, List[int]] = 0,
     ):
-        savedir = os.path.join(savedir, "mlflow")
+        savedir = Path(savedir) / "mlflow"
```
annaelisalappe, line 512: Same here

src/itwinai/loggers.py
```diff
@@ -604,7 +604,7 @@ def log(
             if not isinstance(item, str):
                 # Save the object locally and then log it
                 name = os.path.basename(identifier)
-                save_path = Path(self.savedir) / ".trash" / name
+                save_path = Path(self.savedir) / ".trash" / str(name)
```
annaelisalappe, line 607: And here

src/itwinai/loggers.py
```diff
@@ -685,12 +685,12 @@ class WandBLogger(Logger):
...
     def __init__(
         self,
-        savedir: str = "mllogs",
+        savedir: Path = "mllogs",
         project_name: str = BASE_EXP_NAME,
         log_freq: Union[int, Literal["epoch", "batch"]] = "epoch",
         log_on_workers: Union[int, List[int]] = 0,
     ) -> None:
-        savedir = os.path.join(savedir, "wandb")
+        savedir = Path(savedir) / "wandb"
```
annaelisalappe, line 693: Okay you get it 😅

annaelisalappe (commented):
Love this, the Paths had been annoying me as well! Did you check that this would not break compatibility in places where the loggers are instantiated, like in the use cases?

src/itwinai/loggers.py
```diff
@@ -604,7 +604,7 @@ def log(
             if not isinstance(item, str):
                 # Save the object locally and then log it
                 name = os.path.basename(identifier)
```
jarlsondre, line 606: Regarding this whole function: 

It seems like we do a bunch of pattern matching on kind and item. This might be streamlined by using 

```python
match (kind, item):
            case ("metric", _):
                mlflow.log_metric(key=identifier, value=item, step=step)

            case ("artifact", str()):
                mlflow.log_artifact(local_path=item, artifact_path=identifier)

            case ("artifact", _):
                name = os.path.basename(identifier)
                save_path = Path(self.savedir) / ".trash" / str(name)
                save_path.mkdir(os.path.dirname(save_path), exist_ok=True)
                item = self.serialize(item, save_path)
                mlflow.log_artifact(local_path=item, artifact_path=identifier)
                ...
```
Not super important, but we lose a bit of indentation in which case, making the code a bit more readable.

src/itwinai/loggers.py
```diff
@@ -707,10 +707,8 @@ def create_logger_context(self, rank: Optional[int] = None) -> None:
         if not self.should_log():
             return
 
-        os.makedirs(os.path.join(self.savedir, "wandb"), exist_ok=True)
-        self.active_run = wandb.init(
-            dir=os.path.abspath(self.savedir), project=self.project_name
-        )
+        os.makedirs(self.savedir / "wandb", exist_ok=True)
```
jarlsondre, line 710: In this case you can actually do 

```python
(self.savedir / "wandb").mkdir(exist_ok=True)
```
and avoid `os` altogether.

src/itwinai/loggers.py
matbun: here it could also be a string. Fixing type hint

src/itwinai/loggers.py
matbun: could be str as well

src/itwinai/loggers.py
matbun: yes here I removed it

src/itwinai/loggers.py
matbun: also here it could be a str

src/itwinai/loggers.py
matbun: cool!

src/itwinai/loggers.py
matbun: cool, I missed this feature!

matbun: tried it, didn't like it for complex conditions, e.g., `not isinstance(item, str)`

jarlsondre: I haven't used pattern matching too much in Python, so for the exercise I wrote out what I think the code would look like. I think you can stay at one indentation level if you do this, though I haven't actually tested the code: 

```python
match (kind, item): 
            case ("metric", _): 
                mlflow.log_metric(key=identifier, value=item, step=step)
            case ("artifact", str()): 
                mlflow.log_artifact(local_path=item, artifact_path=identifier)
            case ("artifact", _): 
                name = os.path.basename(identifier)
                save_path = Path(self.savedir) / ".trash" / name
                save_path.mkdir(os.path.dirname(save_path), exist_ok=True)
                item = self.serialize(item, save_path)
                mlflow.log_artifact(local_path=item, artifact_path=identifier)
            case ("model", item) if isinstance(item, torch.nn.Module): 
                import torch
                mlflow.pytorch.log_model(item, identifier)
            case ("model", _): 
                import torch
                print("WARNING: unrecognized model type")
            case ("dataset", item) if isinstance(item, mlflow.data.Dataset): 
                mlflow.log_input(item)
            case ("dataset", _): 
                print("WARNING: unrecognized dataset type. " "Must be an MLFlow dataset")
            case ("torch", _): 
                import torch

                # Save the object locally and then log it
                name = os.path.basename(identifier)
                save_path = Path(self.savedir) / ".trash" / name
                save_path.mkdir(os.path.dirname(save_path), exist_ok=True)
                torch.save(item, save_path)
                # Log into mlflow
                mlflow.log_artifact(local_path=save_path, artifact_path=identifier)
            case ("dict", _): 
                mlflow.log_dict(dictionary=item, artifact_file=identifier)
            case ("figure", _): 
                mlflow.log_figure(
                    artifact_file=identifier,
                    figure=item,
                    save_kwargs=kwargs.get("save_kwargs"),
                )
            case ("image", _): 
                mlflow.log_image(artifact_file=identifier, image=item)
            case ("param", _): 
                mlflow.log_param(key=identifier, value=item)
            case ("text", _): 
                mlflow.log_text(artifact_file=identifier, text=item)
            case _: 
                pass 

```

src/itwinai/loggers.py
```diff
@@ -349,11 +349,11 @@ class ConsoleLogger(Logger):
 
     def __init__(
         self,
-        savedir: str = "mllogs",
+        savedir: Path | str = "mllogs",
         log_freq: Union[int, Literal["epoch", "batch"]] = "epoch",
         log_on_workers: Union[int, List[int]] = 0,
     ) -> None:
-        savedir = savedir = Path(savedir) / "simple-logger"
+        savedir = Path(savedir) / "simple-logger"
```
jarlsondre, line 356: This is fine syntactically, but I am at slight unease with the idea of setting `savedir` to something else. Not very important, but could be nice to use a different variable name instead of updating it. This is also true for some of the other functions in this file.

src/itwinai/loggers.py
```diff
@@ -1031,9 +1032,9 @@ class Prov4MLLogger(Logger):
 
     def __init__(
         self,
-        prov_user_namespace="www.example.org",
-        experiment_name="experiment_name",
-        provenance_save_dir="mllogs/prov_logs",
+        prov_user_namespace: str = "www.example.org",
+        experiment_name: str = "experiment_name",
+        provenance_save_dir: Path = "mllogs/prov_logs",
```
jarlsondre, line 1037: You set the type to `Path` but the default argument is a `str`.

src/itwinai/loggers.py
```diff
@@ -113,7 +113,7 @@ class Logger(LogMixin):
     """Base class for logger
 
     Args:
-        savedir (str, optional): filesystem location where logs are stored.
+        savedir (Union[Path, str], optional): filesystem location where logs are stored.
```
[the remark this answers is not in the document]
matbun: reverted back to `Union[path, str]`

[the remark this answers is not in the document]
matbun: the `savedir` property of the Logger class is of type `Path`, but the constructor argument can also be a string

src/itwinai/loggers.py
matbun: yes, this made me notice that the constructor's `savedir` does not coincide with the homonymous class attribute. This would need some refactor, but it's a bit out of the scope of this PR. nice comment though

src/itwinai/loggers.py
```diff
@@ -240,7 +239,7 @@ def serialize(self, obj: Any, identifier: str) -> str:
         Returns:
             str: local path of the serialized object to be logged.
         """
-        itm_path = os.path.join(self.savedir, identifier)
+        itm_path = self.savedir / str(identifier)
```
jarlsondre, line 242: "identifier" should already be of type "str" here.

matbun: true

jarlsondre (dismissed):
Pathlib ❤️

jarlsondre (approved):

PR #262, opened 2024-12-02
Update Scalability Tutorial

jarlsondre:
# Summary
This PR updates the scalability tutorial in multiple ways: 
- Cleaner code
  - Removing many prints that strictly were not necessary
  - Reusing functionality such as the argparser or the `train_epoch` function for easier readability
  - General formatting
  - Rewritten tutorial
- SLURM scripts
  - Uses the new Python scripts for running the SLURM jobs, e.g. running all the strategies and also the scaling test

# Current Limitations
The scalability tutorial was traditionally created with simplicity in mind, and therefore it tries to minimize its dependency to the `itwinai` project. Because of this, it does not use the `itwinai` trainer, which is required for using the updated scalability metrics. Therefore, the current implementation of the scalability tutorial only has the absolute and relative wall-clock time plots. This is, however, not so different from how it was before. 

# Examples of resulting plots
This is done using a subset size of 5000, so not the full ImageNet dataset. 

## Relative Scalability Plot
image

## Absolute Scalability Plot
image

# List of tasks
I am including this list of tasks not only for my own benefit, but also to make it easier for reviewers to see what has been done. Keep in mind that this list is not exhaustive: I started adding to it after already finishing some tasks. 
- [x] Fix deepspeed local rank problem
  - Turned out to be caused by some environment variables related to OMPI local rank 
- [x] Fix Horovod not wanting to load stuff (runs out of time?)
  - This turned out to be caused by me forgetting to train on a _subset_ of imagenet instead of the whole thing...  
- [x] Change the name of the default SLURM log dir
- [x] Refactor the tutorial files
  - No need to make them perfect but there is a lot of redundant stuff that just causes confusion IMO
- [x] Update the tutorial readme
- [x] Allow user to specify nodes for scalability analysis
- [x] Allow user to specify imagenet subset size
- [x] Allow use to specify folder for storing scalability metrics

**Related issue :**
#221, #263


tutorials/distributed-ml/torch-tutorial-1-mnist/train.py
```diff
@@ -44,7 +44,7 @@ def parse_params() -> argparse.Namespace:
 
     >>> train.py --strategy ddp --config config.yaml
     """
-    parser = ItAIArgumentParser(description="PyTorch MNIST Example")
+    parser = ItwinaiArgParser(description='PyTorch MNIST Example')
```
matbun, line 47: why apostrophes and not quotation marks?

use-cases/virgo/config.yaml
```diff
@@ -48,17 +48,17 @@ training_pipeline:
...
+          #   class_path: itwinai.loggers.LoggersCollection
+          #   init_args:
+          #     loggers:
+          #       - class_path: itwinai.loggers.ConsoleLogger
+          #         init_args:
+          #           log_freq: 1
+          #       - class_path: itwinai.loggers.MLFlowLogger
+          #         init_args:
+          #           experiment_name: Noise simulator (Virgo)
+          #           log_freq: batch
```
matbun, line 61: is there any reason to disable the loggers?

uv-tutorial.md
matbun, line 1: could you move this file under `docs/`, pls?

tutorials/distributed-ml/torch-tutorial-1-mnist/train.py
```diff
@@ -25,7 +25,7 @@
 from torch.utils.data import Dataset
 from torchvision import datasets, transforms
 
-from itwinai.parser import ArgumentParser as ItAIArgumentParser
```
annaelisalappe, line 28: Do you think this could in the future be replaced by the new Parser? I didn't know that the ArgumentParser was used anywhere other than in the pipeline execution, but it's definitely good to see. Is there a reason why you need to use jsonargparse here (so basically parsing yaml) or is it just used as a cli parser here?

use-cases/virgo/config.yaml
```diff
@@ -26,7 +26,7 @@ training_pipeline:
     steps:
       - class_path: data.TimeSeriesDatasetSplitter
         init_args:
-          train_proportion: 0.9
+          # train_proportion: 0.0
```
annaelisalappe, line 29: Is this supposed to be commented out?

annaelisalappe (commented):
I'm a little unsure if my build of the docs worked correctly, you might have to show me later where I can find the new tutorial in the docs :)

uv-tutorial.md
jarlsondre: Haha apparently this file is a duplicate file, because it is already moved under `docs/` already. Deleted this file now.

use-cases/virgo/config.yaml
jarlsondre: No this was just done for experimentation purposes. Commented them back in now.

src/itwinai/slurm/slurm_config.yaml
```diff
@@ -3,8 +3,8 @@ account: intertwin
 dist_strat: horovod
 time: 00:11:11
 
-std_out: slurm_jobs/job.out
-err_out: slurm_jobs/err.out
+std_out: slurm_job_Logs/job.out
```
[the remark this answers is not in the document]
jarlsondre: Nope, just a silly mistake. Fixed it now.

tutorials/distributed-ml/torch-tutorial-1-mnist/train.py
jarlsondre: Whether or not it can be replaced by the new parser depends on the capabilities of the parser. If we make our own implementation of a parser it would naturally be nice to use that everywhere. Currently, what I require in this case is to be able to 1) set up the arguments with the known `parser.add_argument()` syntax and 2) read arguments from a configuration file. Right now I don't have to use the "-o" syntax, which I think is nice, so would be nice to not have to use that also. That's not a requirement, though

use-cases/virgo/config.yaml
jarlsondre: No, it is not. Fixed it now.

tutorials/distributed-ml/torch-tutorial-1-mnist/train.py
annaelisalappe: Yeah you're right, we can look into that at another place, definitely not here 😌

annaelisalappe (approved):

Issue #263, opened 2024-12-03
Updated handling of SLURM scripts


jarlsondre:
# Summary
Currently there are a lot of duplicate bash scripts for submitting SLURM jobs. This is inefficient and hard to maintain, so this should be rewritten. As these scripts benefit from reusability and thus slightly more complex scripting, the suggestion is to do this in Python instead. Python is generally better suited for more complex scripting tasks and everyone in this project is familiar with Python. 

Using libraries like PySlurm is possible, but this would create a slight learning curve for anyone not familiar with this library, and it is far from an industry standard. Therefore I propose that we use write the code for this ourselves. This also means that we require one less dependency, which is always an advantage. 

**Wanted features**: 
- Ability to specify your slurm job in a yaml file, so that everything can be changed in a single spot
- Centralized commands that are common to the different use cases

**Ideas (food for thought)**: 
- In each use-case folder there should be a config file that you use to specify your job
- There should be some way to override functions for a specific use case, if need be
- Avoiding over-abstraction is important. The main goal here is a higher level of reusability, not an entire SLURM library.

PR #265, opened 2024-12-03
Slurm script update

jarlsondre:
# Summary
This implements the SLURM scripts in Python instead of bash, for more simplicity in the long run.  

The current implementation allows you to use configuration files to run SLURM jobs, bypassing the need for one (or multiple) bash script(s) to do the same. If you need more fine-grained control, you can also inherit from the `SlurmScriptBuilder`class and write your own logic. This makes it very easy to do things such as running all the strategies or doing scalings tests. 

Below, you can see an example of using the built in builder. Assuming you have a configuration file called `slurm_config.yaml`, you can pass it with `-c` and then add/override any arguments you want after. 

```bash
python slurm_script_builder.py -c slurm_config.yaml --no-submit-job --no-retain-file --debug
```

In this case, we do `--no-submit-job` and `--no-retain-file`, meaning that it will not run the script with sbatch and it will also not save it to file. When both of these are active, the builder will simply print the script to the console. This allows for a quick feedback loop when creating your config, as you can make changes, look at the definitive output without having to wonder what the variables are evaluated to, and repeat until you get exactly the script you want. Afterwards, you can then remove `--no-submit-job` to make sure it is being executed and if you want to store the same script for later (e.g. for reproducibility) then you can also remove `--no-retain-file` to make it automatically save to file. 

# Future Plans
This PR contains the initial design for the SLURM job creator. There are more features that we wish to add, but these might be more suitable for another PR, as I currently have something else that is being blocked by this. Besides, getting your input after testing this out is also valuable. Some features for a future PR are: 
- Integrate this into the CLI tool for itwinai
- Only use a single config file, the one from the use cases, instead of having two separate ones
- Add SLURM scripts for Ray stuff

**Related issue :**
#263


src/itwinai/slurm/slurm_script_builder.py
```diff
@@ -0,0 +1,299 @@
...
+        setup_command = setup_command.strip()
+        return remove_indentation_from_multiline_string(setup_command)
+
+    def get_slurm_script(
+        self, setup_command: str | None = None, main_command: str | None = None
+    ):
+        if setup_command is None:
+            setup_command = self.get_setup_command()
+        if main_command is None:
+            main_command = self.get_srun_command()
```
matbun, line 155: rename as `srun_command` to keep naming consistent

src/itwinai/slurm/slurm_script_builder.py
```diff
@@ -0,0 +1,299 @@
...
+        echo "SLURMD_NODENAME: $SLURMD_NODENAME"
+        echo "CUDA_VISIBLE_DEVICES: $CUDA_VISIBLE_DEVICES"
+        echo ""
+        echo "### Other Variables ###"
+        echo "Distributed Strategy: {self.distributed_strategy}"
+        """
+        debug_print_command = debug_print_command.strip()
+        return remove_indentation_from_multiline_string(debug_print_command)
+
+    def get_setup_command(self):
```
matbun, line 136: I would rename setup as pre_exec for better clarity, and possibly add a post_exec to be run after the srun (e.g., cleanup)

src/itwinai/slurm/slurm_script_builder.py
```diff
@@ -0,0 +1,299 @@
...
+            # Resources allocation
+            #SBATCH --nodes={self.num_nodes}
+            #SBATCH --ntasks-per-node={self.num_tasks_per_node}
+            #SBATCH --gpus-per-node={self.gpus_per_node}
+            #SBATCH --cpus-per-gpu={self.cpus_per_gpu}
+            #SBATCH --exclusive
+
+            {setup_command}
+
+            {main_command}"""
```
matbun, line 178: you can save this template as a constant and extract it from this class

See `from string import Template`

src/itwinai/slurm/slurm_script_builder.py
```diff
@@ -0,0 +1,299 @@
...
+        echo "SLURM_NNODES: $SLURM_NNODES"
+        echo "SLURM_NTASKS: $SLURM_NTASKS"
+        echo "SLURM_TASKS_PER_NODE: $SLURM_TASKS_PER_NODE"
+        echo "SLURM_SUBMIT_HOST: $SLURM_SUBMIT_HOST"
+        echo "SLURMD_NODENAME: $SLURMD_NODENAME"
+        echo "CUDA_VISIBLE_DEVICES: $CUDA_VISIBLE_DEVICES"
+        echo ""
+        echo "### Other Variables ###"
+        echo "Distributed Strategy: {self.distributed_strategy}"
+        """
```
matbun, line 132: I'd probably extract this from the class as well

src/itwinai/slurm/slurm_script_builder.py
matbun, line 1: Maybe interesting to look at Builder pattern?

src/itwinai/slurm/slurm_config.yaml
matbun, line 1: is this file meant to be here?

src/itwinai/slurm/slurm_constants.py
matbun, line 1: add credits

src/itwinai/slurm/slurm_constants.py
```diff
@@ -0,0 +1,34 @@
+slurm_template = r"""#!/bin/bash
```
matbun, line 1: since it's a "constant" I would maybe give it a capitalized name

src/itwinai/slurm/slurm_script_builder.py
```diff
@@ -0,0 +1,331 @@
...
+        training_command=args.training_cmd,
+    )
+
+    slurm_script_builder.process_slurm_script(
+        retain_file=retain_file, submit_slurm_job=submit_slurm_job
+    )
+
+
+if __name__ == "__main__":
+    main()
```
matbun, line 331: it's fine to leave this here, but I was wondering whether you forgot to remove this. Maybe this could be converted in unit tests?

src/itwinai/slurm/slurm_config.yaml
jarlsondre: I intended for this to serve as a placeholder/example for using the general SLURM builder, but plan on removing/moving it once I implement the CLI stuff.

matbun (approved):

PR #276, opened 2024-12-13
Add more Search Algorithms and Schedulers for Tuning

annaelisalappe:
# Summary

This PR adds all of the search algorithms and schedulers supported by Ray Tune, and refactors the script where they are parsed and set. 
It also refactors how the configuration is handled in the RayTorchTrainer. I decided that it was best to convert the initial dictionary to a `TrainingConfiguration` from within the` train()` function, because even defining new `Dict` - like classes always threw errors - I also think that this way is the most intuitive, conforms with Ray standards which makes it easier to debug, yet still gives us the type-checking advantages from converting into a `TrainingConfiguration`. 

There are very thorough error messages when the right fields for some things, such as the Ray Configs or the Search Algorithms and Schedulers are not set - I found this to be the least convoluted way of making sure the user can set all relevant fields if they want to, yet not blowing up the code significantly by parsing each one manually from the dictionary. 

I will have a look at the logging and reporting of metrics in the next PR 😌


src/itwinai/torch/trainer.py
```diff
@@ -1429,130 +1400,106 @@ def set_epoch(self, epoch: int) -> None:
...
 
-            elif param_type == "randn":
-                train_loop_config[name] = tune.randn(values["mean"], values["stddev"])
+    def _set_train_loop_config(self) -> None:
+        train_loop_config = self.config.get("train_loop_config", {})
 
-            elif param_type == "qrandn":
-                train_loop_config[name] = tune.qrandn(
-                    values["mean"], values["stddev"], values["q"]
+        if train_loop_config:
```
jarlsondre, line 1480: If you handle the else part first, which is just a print statement followed by (I believe) returning an empty dict, you can remove a lot of indentation here.

src/itwinai/torch/trainer.py
```diff
@@ -1429,130 +1400,106 @@ def set_epoch(self, epoch: int) -> None:
...
+            try:
+                for name, param in train_loop_config.items():
+                    if isinstance(param, dict):
```
jarlsondre, line 1483: If you negate this and add a continue, you can remove a bit of indentation here.

src/itwinai/torch/tuning.py
```diff
@@ -1,18 +1,44 @@
...
+    TuneBOHB
+    | BayesOptSearch
+    | HyperOptSearch
+    | AxSearch
+    | HEBOSearch
+    | NevergradSearch
+    | OptunaSearch
+    | ZOOptSearch
+    | None
+):
```
jarlsondre, line 41: It might be easier to annotate this as "Searcher" (the super class for the search algorithms" instead of mentioning every single one, but not something you have to change.

src/itwinai/torch/tuning.py
```diff
@@ -23,59 +49,63 @@ def get_raytune_search_alg(
...
-        )
-
-    # requires pip install hyperopt
-    if search_alg == "hyperopt":
-        print("INFO: Using HyperOptSearch")
-        return HyperOptSearch(
-            n_initial_points=tune_config["search_alg"]["n_random_steps"],
-            # points_to_evaluate=,
+    try:
+        match search_alg_name.lower():
```
jarlsondre, line 76: wow 👀❤️

src/itwinai/torch/tuning.py
```diff
@@ -89,48 +119,36 @@ def get_raytune_schedule(
...
-            time_attr="training_iteration",
-            perturbation_interval=tune_config["scheduler"]["perturbation_interval"],
-            hyperparam_bounds=tune_config["scheduler"]["hyperparam_bounds"],
-            log_config=True,
+
+    scheduler = tune_config.pop("scheduler", {})
+    scheduler_name = scheduler.pop("name", "")
+
+    try:
+        match scheduler_name.lower():
```
jarlsondre, line 133: hot damn

src/itwinai/torch/tuning.py
annaelisalappe: 🤣

src/itwinai/torch/tuning.py
annaelisalappe: I get what you mean! I actually like having them listed like this to quickly see which algorithms are implemented. Although, to be fair, it's not exactly hard to see from the code

jarlsondre: I think that either is okay. The good thing about keeping it verbose like this is that you can see all the implemented strategies, the downside is that you have to remember to update this if you add another strategy. I'd say just do whichever you prefer :)

jarlsondre (approved):
woop woop

Issue #172, opened 2024-06-25
Kubeflow tutorial


matbun:
Tutorial on Kubeflow (pipelines, models registry, tune...)

PR #280, opened 2025-01-07
Distributed ml on k8s - first tutorial

matbun:
This is somehow related to #172, but introduces distributed ML on k8s using Kubeflow's training operator, rather than focusing on pipelines or ML tracking. 

Future work include more work on GPU nodes and the inclusion of persistent volumes.


docs/getting-started/slurm.rst
```diff
@@ -65,7 +65,18 @@ Once resources are available, the command will return a ``JOBID``. Use it to jum
    # Check that you are in the compute node and you have 4 GPUs
    nvidia-smi
 
-Eventually, remember to load the correct environment modules before activating the python virtual environment.
+Remember to load the correct environment modules before activating the python virtual environment.
+
+Alternatively, if you don't need to open a terminal on the compute node allocated with the
```
jarlsondre, line 70: This feels a little bit unclear to me. Do you mean that if it is okay for the program to run in the background and you don't need access to the terminal, then you can use `srun` instead of `salloc`?

docs/tutorials/tutorials.rst
```diff
@@ -21,6 +21,7 @@ Distributed ML with PyTorch
    distrib-ml/torch-tutorial-GAN
    distrib-ml/torch_scaling_test
    distrib-ml/torch-tutorial-containers
+   distrib-ml/torch_tutorial_kubeflow_1.rst
```
jarlsondre, line 24: This is probably fine, but I am just curious about the naming convention here: Why do you add "_1" at the end?

src/itwinai/torch/distributed.py
```diff
@@ -154,15 +154,18 @@ def device(self) -> str:
...
 
     def set_device(self):
         """Set local device."""
-        torch.cuda.device(self.local_rank())
-        # Needed by torch.distributed.gather_object
-        torch.cuda.set_device(self.local_rank())
+        if torch.cuda.is_available():
+            torch.cuda.device(self.local_rank())
+            # Needed by torch.distributed.gather_object
+            torch.cuda.set_device(self.local_rank())
```
jarlsondre, line 168: I know that this was here already, but why is it that we use `set_device()` here instead of just handling this using the `.to(device)` syntax? The reason I ask is because of this description in the [PyTorch Documentation](https://pytorch.org/docs/stable/generated/torch.cuda.set_device.html): 

> Set the current device.
> 
> Usage of this function is discouraged in favor of [device](https://pytorch.org/docs/stable/generated/torch.cuda.device.html#torch.cuda.device). In most cases it’s better to use CUDA_VISIBLE_DEVICES environmental variable.

In particular, I am thinking of the "this function is discouraged in favor of device" statement. Maybe this is only for the general case and `set_device()` is better for distributed learning?

tutorials/distributed-ml/torch-kubeflow-1/train-cpu.py
```diff
@@ -0,0 +1,162 @@
...
+
+    transform = transforms.Compose(
+        [transforms.ToTensor(), transforms.Normalize((0.1307,), (0.3081,))]
+    )
+
+    is_main_worker = (
+        not os.environ.get("LOCAL_RANK")
+        or os.environ.get("LOCAL_RANK")
+        and os.environ["LOCAL_RANK"] == "0"
+    )
```
jarlsondre, line 93: This is somewhat confusing due to the extensive use of `.get()`, so might be easier to store this in a variable and then do an if-else. Not a huge deal since the variable name is pretty self-explanatory, but I had to stop for a second to understand the logic here.

jarlsondre: It's okay the way it is

tutorials/distributed-ml/torch-kubeflow-1/train-cpu.py
```diff
@@ -0,0 +1,162 @@
...
+    if is_main_worker:
+        # Dataset creation: pull dataset only in main worker
+        train_dataset = datasets.MNIST(
+            "data",
+            train=True,
+            download=not os.path.exists("DATASET_READY"),
```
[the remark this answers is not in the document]
jarlsondre: It's okay the way it is.

docs/getting-started/slurm.rst
matbun: I have rephrased it. I hope it's more clear now. If not, could you suggest a better formulation, pls?

jarlsondre: I suppose I'd say something along the lines of "If you don't need to access the compute node's terminal then you can run your command using `srun` instead of `salloc`. Example:"

I think the phrase that messed with my head was "if you don't need to open a terminal on the compute node", haha, but I suppose my pedanticism gets the best of me sometimes. It's probably fine the way it is.

jarlsondre: I see what you mean. My bad, I was only reading the changed part, but when you put it into context I agree with what you're saying--it's fine the way it is. Sorry for the misunderstanding.

jarlsondre (approved):
LGTM

PR #282, opened 2025-01-08
[DRAFT] Config Parser

annaelisalappe:
# Summary

The main contribution of this PR is the change from the old ConfigParser (which was based on jsonargparse) to an execution based on hydra. The advantage of this change is that error messages that arise during execution of the pipeline were often not visible in the error message given out by the jsonargparse parser, whereas in hydra they are passed on. This makes debugging much easier. An example of the difference:

If I set the seed in the config file to "small", I get the following outputs from the old parser and the new one, respectively:

Screenshot 2025-01-08 at 16 50 20

Screenshot 2025-01-08 at 16 51 42


src/itwinai/cli.py
```diff
@@ -157,85 +165,66 @@ def sanity_check(
...
-        pipeline = pipeline[steps]
-    pipeline.execute()
+    # Process CLI arguments to handle paths
+    sys.argv = make_config_paths_absolute(sys.argv)
+
+    exec_pipeline_with_compose()
+
+
+@hydra.main(version_base=None, config_path=os.getcwd(), config_name="config")
+def exec_pipeline_with_compose(cfg):
```
jarlsondre, line 189: I don't love the way the `cfg` parameter is passed by the decorator here. At this point, you can't really know that this is what happens without knowledge of exactly how the `hydra` decorator works. In other words, there is a bit of hidden handling here, that hurts the readability. 

The most naive solution to this would be to add it in the docstring/comments. However, I can see that there is something called `initialize` and `compose` from `hydra` that might allow you to achieve the same thing without the hidden decorator functionality: https://hydra.cc/docs/advanced/compose_api/

src/itwinai/cli.py
```diff
@@ -157,85 +165,66 @@ def sanity_check(
...
+        print("No steps selected. Executing the whole pipeline.")
+
+    # Instantiate and execute the pipeline
+    try:
+        pipeline = instantiate(cfg, _convert_="all")
+        pipeline.execute()
+    except Exception as e:
+        root = get_root_cause(e)
+        raise e
+        raise root
```
jarlsondre, line 227: I'd say pick `raise e` or `raise root`, but maybe not both ;)

src/itwinai/serialization.py
```diff
@@ -84,7 +84,10 @@ def to_dict(self) -> Dict:
         init_args = dict()
         for par_name, par in self._saved_constructor_parameters().items():
             init_args[par_name] = self._recursive_serialization(par, par_name)
-        return dict(class_path=class_path, init_args=init_args)
+
+        init_args["_target_"] = class_path
+
```
jarlsondre, line 89: This empty line bothers me more than it should hahaha

src/itwinai/slurm/utils.py
```diff
@@ -127,10 +127,16 @@ def get_slurm_job_parser() -> ArgumentParser:
...
+        "--config-path",
         type=str,
         default=default_config_file,
-        help="Which config file to use for training.",
+        help="The path to the directory containing the config file to use for training.",
+    )
+    parser.add_argument(
+        "--config-name",
+        type=str,
+        default=default_config_file,
```
jarlsondre, line 138: Should the default be the same as the above argument here?

src/itwinai/utils.py
```diff
@@ -184,6 +159,34 @@ def clear_key(my_dict: Dict, dict_name: str, key: Hashable, complain: bool = Tru
...
         if complain:
-            print(f"Field '{key}' should not be present " f"in dictionary '{dict_name}'")
+            print(f"Field '{key}' should not be present in dictionary '{dict_name}'")
         del my_dict[key]
     return my_dict
+
+
+def make_config_paths_absolute(args):
+    """Process CLI arguments to make paths specified for `--config-path` or `-cp` absolute.
+    Returns the modified arguments list."""
```
jarlsondre, line 169: You should add some type hints here. I would add the `: <type>` ones anyway, but you can also considering adding some extra explanations in the docstring

src/itwinai/cli.py
annaelisalappe: Yes you're right, but by using `initialize` we loose basically all of the benefits of using hydra - the idea with this whole thing was really to use the hydra implementation as much as-is as possible, because hydra handles a lot of things, e.g. debugging, cli options for overriding parameters, etc. You loose all of these by just using initialize, then we would have to write the `exec-pipeline` function to have cli args for all of these, and manually pass them to hydra `initialize`. This is possible, but not really the way you are supposed to use hydra at all, it's more made as a workaround in case you have to call hydra from within your code, like it is done in the `hpo.py` scripts. You can read more about the hydra decorator here: https://hydra.cc/docs/tutorials/basic/your_first_app/simple_cli/. So basically, we are using the decorator here because that is how you are supposed to use hydra - not using it would sort of defy the purpose of using it in the first place :)

src/itwinai/cli.py
annaelisalappe: Okay but the point here was to stack the errors together, so you have the root cause in front of the full traceback for easier debugging :) At least that's how I remembered it from the conversation we had with Matteo, correct me if I'm wrong!

src/itwinai/slurm/utils.py
annaelisalappe: No it shouldn't, good catch!

src/itwinai/utils.py
```diff
@@ -184,6 +159,34 @@ def clear_key(my_dict: Dict, dict_name: str, key: Hashable, complain: bool = Tru
...
+            updated_args[i] = f"{prefix}={abs_path}"
+            sys.path.append(abs_path)
+            break
+        elif arg in {"--config-path", "-cp"}:
+            # Handle the case where the path is in the next argument
+            abs_path = os.path.abspath(updated_args[i + 1])
+            updated_args[i + 1] = abs_path
+            sys.path.append(abs_path)
+            break
+    return updated_args
```
jarlsondre, line 184: First of all: `pathlib` 

But I was thinking about this function. I don't know which type `args` is here (😉), but it seems to be a list. This method would probably be a whole lot easier if the input was a dictionary, so I wondering: Will this ever turn into a dictionary later? Because in which case, maybe you could call this function later and make it a lot shorter? 

Assuming `args` becomes a dict later (which it might not), then you could just do
```python
args["config_path"] = Path(args["config_path"]).resolve()
```
or something along those lines. It would remove the need for as much indexing and splitting, which inherently are bug-prone instructions. 

Also, if you don't want to do that, then you still could move the `sys.path.append(abs_path)` to after the for loop, since you want to do this regardless. I suppose that could be problematic if you can't guarantee that `--config-path` actually occurs in `args` though. Maybe you know if this is the case?

annaelisalappe: Okay haha, I'll try to go through one-by-one 😃

1. Yes `pathlib`, but I actually left it like this intentionally, because when you append to os.sys.path (at least afaik), you have to convert it back to a string - so I would do 
```
abs_path = Path(path).resolve()
args[i] = f"{prefix}={abs_path}"
sys.path.append(str(abs_path))
```.
I could of course, but I found the first one neater.

2. No it is never converted to a dict - args is the list of command-line arguments, so it is always staying as a list :) I updated the doctoring to make it clearer!

3. I can't move `sys.path.append(abs_path)` down because it could indeed not be in the list, if the user doesn't want to override the default path.

File: src/itwinai/cli.py (hunk 1 above)
Comment (jarlsondre) [reply src/itwinai/cli.py:189]: I see. I agree then that we should keep the implementation as it is. We should add a note in the docstring though, because this threw me off for soooo long the first time I saw it hahaha

File: src/itwinai/cli.py (hunk 2 above)
Comment (jarlsondre) [reply src/itwinai/cli.py:227]: I see what you mean, but I'm fairly certain (though could be wrong) that you can't raise multiple exceptions. That is to say, once `raise e` is evaluated, the code "leaves the function" and `raise root` is never called.

File: src/itwinai/utils.py (hunk 6 above)
Comment (jarlsondre) [reply src/itwinai/utils.py:184]: I see, thanks for clarifying. 

1) I agree, using `os.path.abspath` is fine here since we only use a single function and don't have to deal with paths for a prolonged amount of time. Additionally, like you're saying, you'd have to turn it back into a `str`. 

2) Aha, then it makes sense to me. I guess I was mixing this up with the config file or something. 

3) I see, yep, makes sense.

File: tests/components/test_config_parsing.py [hunk 7]
```diff
@@ -0,0 +1,187 @@
...
+        "my-dict-pipeline": pytest.PIPE_DICT_YAML,
+        "some.field.my-nested-pipeline": pytest.NESTED_PIPELINE,
+        "my-interpolation-pipeline": pytest.INTERPOLATED_VALUES_PIPELINE,
+    }
+
+    file_paths = {}
+
+    for name, content in yaml_data.items():
+        file_path = tmp_path / f"{name.lower()}.yaml"
+        with file_path.open("w") as f:
```
jarlsondre, line 37: Might not this fail if the file_path does not exist? Not sure how important this is, considering it is a test, but just something I started thinking about.

src/itwinai/cli.py
annaelisalappe: Oh okay, good point 😅 I think then raising just `e` is actually the best way. The output of `e` is (most of the time, as far as I can tell) only a very short version and basically just the original error propagated through. If you want a more complete error message, you can set `HYDRA_FULL_ERROR=1`, which the error message also tells you.

tests/components/test_config_parsing.py
```diff
@@ -0,0 +1,187 @@
...
+
+                exec_pipeline()
+
+                mock_preproc_init.assert_called_once_with(max_items=33, name="my-preproc")
+                mock_preproc_exec.assert_called_once()
+
+                mock_trainer_init.assert_called_once_with(
+                    lr=0.001, batch_size=32, name="my-trainer"
+                )
+                mock_trainer_exec.assert_called_once()
```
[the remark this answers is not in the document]
annaelisalappe: These are defined in `conftest.py` in that same folder, there I defined some pipelines for testing :) So it tests whether the values are what they are supposed to be, of correct type, etc.

tests/use-cases/test_3dgan.py
```diff
@@ -39,14 +40,14 @@ def test_3dgan_train(torch_env, install_requirements):
     """
     install_requirements(CERN_PATH, torch_env)
     dataset_path = os.environ.get("CERN_DATASET", DEFAULT_DATASET_PATH)
-    conf = (CERN_PATH / "config.yaml").resolve()
+    conf = CERN_PATH.resolve()
```
jarlsondre, line 43: Maybe `CERN_PATH` could just be resolved when it is defined instead of having to resolve it everywhere it is used? There might not be any reason to keep it relative?

tests/components/test_config_parsing.py
annaelisalappe: Yes this would probably fail, but I would argue that is fine in a test. We create the file_paths ourselves in the same test run, so it should be fine.

tests/use-cases/test_mnist.py
```diff
@@ -107,11 +106,11 @@ def test_mnist_train_torch_lightning(torch_env, install_requirements):
     install_requirements(LIGHTNING_PATH, torch_env)
 
     dataset_path = os.environ.get("MNIST_DATASET", DEFAULT_MNIST_DATASET)
-    conf = (LIGHTNING_PATH / "config.yaml").resolve()
+    conf = LIGHTNING_PATH.resolve()
```
jarlsondre, line 109: Same as with the `CERN_PATH´ above (resolve etc.)

use-cases/3dgan/trainer.py
```diff
@@ -84,6 +84,8 @@ def _log_config(self, logger: Logger):
         with tempfile.TemporaryDirectory(dir="/tmp") as tmp_dir:
             local_yaml_path = os.path.join(tmp_dir, "pl-conf.yaml")
             with open(local_yaml_path, "w") as outfile:
+                print(f"Config: {self.conf}")
+                print(f"Type of config: {type(self.conf)}")
```
jarlsondre, line 88: These are only here for debugging, right?

use-cases/eurac/slurm.sh
```diff
@@ -31,15 +31,20 @@ if [ -z "$RUN_NAME" ]; then
   >&2 echo "WARNING: env variable RUN_NAME is not set. Defaulting to $DIST_MODE."
   RUN_NAME=$DIST_MODE
 fi
-if [ -z "$CONFIG_FILE" ]; then 
-  CONFIG_FILE=config.yaml
+if [ -z "$CONFIG_PATH" ]; then 
+  CONFIG_PATH=.
   >&2 echo "WARNING: env variable CONFIG_FILE is not set. Defaulting to $CONFIG_FILE."
```
jarlsondre, line 36: I'm hoping to delete this file in the future anyway, but shouldn't it be `CONFIG_PATH` here instead of `CONFIG_FILE`?

tests/use-cases/test_3dgan.py
annaelisalappe: Yeah you're probably right 👍🏻

src/itwinai/cli.py
```diff
@@ -157,85 +165,62 @@ def sanity_check(
...
+            raise e
+    else:
+        print("No steps selected. Executing the whole pipeline.")
+
+    # Instantiate and execute the pipeline
+    try:
+        pipeline = instantiate(cfg, _convert_="all")
+        pipeline.execute()
+    except Exception as e:
+        raise e
```
jarlsondre, line 223: Since you're just re-raising `e` here, you might as well just do: 

```python
pipeline = instantiate(cfg, _convert_="all")
pipeline.execute()
```
That is, you can remove the `try-except` block, as it doesn't do anything at this point.

tests/use-cases/test_mnist.py
```diff
@@ -61,11 +62,11 @@ def test_mnist_train_torch(torch_env, install_requirements):
 
     dataset_path = os.environ.get("MNIST_DATASET", DEFAULT_MNIST_DATASET)
 
-    conf = (TORCH_PATH / "config.yaml").resolve()
+    conf = TORCH_PATH.resolve()
```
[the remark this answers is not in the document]
annaelisalappe: Added this for all three!

jarlsondre (approved):
Cool! Don't know if @matbun wants to have a look before merging or not, but LGTM!

jarlsondre (dismissed):
Looks good to me

matbun (commented):
Great PR!!

Could you please check that the tutorials on "machine learning workflows" are up to date with the latest changes? 
https://itwinai.readthedocs.io/latest/tutorials/tutorials.html#machine-learning-workflows

(Apologies for the late reply)

matbun (approved):

Issue #283, opened 2025-01-09
Loggers refactoring


matbun:
Improvements to the loggers

- [ ] Extract Logger and LoggerMixin from the loggers file, to avoid loading all the dependencies of all logging frameworks when you just need the abstract base class or the mixin
- [ ] Simplify `log_freq`, splitting it into `log_freq_epoch` (int) and `log_freq_batch` (int). In both cases freq=0 means no logging.
- [ ]  Refine common structure (e.g., save_dir)

@annaelisalappe please add other things you had in mind

PR #285, opened 2025-01-10
Lighten logger impact on installation and dynamic import

### LINKED ISSUE #278: Make Prov4ML an optional dependency
the provenance logger has many dependencies, which slow down and make the installation heavier, but has been rarely used so far.
Moreover, it is somewhat unstable and it would be better to have it as an optional installation dep.

matbun:
# Summary

(Hopefully) solves [this](https://github.com/interTwin-eu/itwinai/actions/runs/12706173209/job/35418602174) failing job, since PyPI does not support repositories as direct dependencies (e.g., `prov4ml@git+https://github.com/matbun/ProvML@new-main`).

Moreover, reduced the overhead of dynamic import of `itwinai.loggers` by moving the import of logging frameworks (e.g., mlflow) inside the dedicated itwinai wrapper. This also benefits all the code that depends on `itwinai.loggers` (e.g., `itwinai.torch.trainer`), reducing import time.

Also, moved the `EmptyLogger` at the bottom of `itwinai.loggers` module.

I also took the chance to simplify the imports in the `itwinai.torch.trainer` module.

From a simple experiment I ran on Vega's login node, simplifying imports reduced import time of `TorchTrainer` from ~26.6 to ~0.6 seconds:

```python
start = time.time()
from itwinai.torch.trainer import TorchTrainer
print(f"time spent importing itwinai: {time.time() - start:.6f}")
``` 

I don't know how relevant/general this result is, but it seems that import time has been reduced.

---

**Related issue :**

Related to #283 
Closes #278


src/itwinai/torch/trainer.py
```diff
@@ -614,31 +604,31 @@ def train_epoch(self, epoch: int) -> torch.Tensor:
...
+            train_loss_sum += loss
+            for name, val in metrics.items():
+                train_metrics_sum[name] += val
 
             # Important: update counter
             self.train_glob_step += 1
 
         # Aggregate and log losses
-        avg_loss = torch.mean(torch.stack(train_losses))
+        avg_loss = train_loss_sum.item() / batch_counter
```
jarlsondre, line 615: I might be mistaken here, but it seems to me like `batch_counter` is never updated from zero and thus remains zero for the division. Is this the case or am I missing something? :))

annaelisalappe (approved):
Looks god to me :)

src/itwinai/torch/trainer.py
```diff
@@ -701,29 +691,30 @@ def validation_epoch(self, epoch: int) -> Optional[torch.Tensor]:
...
+            validation_loss_sum += loss
+            for name, val in metrics.items():
+                validation_metrics_sum[name] += val
 
             # Important: update counter
             self.validation_glob_step += 1
 
         # Aggregate and log losses
-        avg_loss = torch.mean(torch.stack(validation_losses))
+        avg_loss = validation_loss_sum.item() / batch_counter
```
jarlsondre, line 707: Same as above with the `batch_counter`

src/itwinai/torch/trainer.py
```diff
@@ -1412,7 +1417,9 @@ def execute(
             Tuple[Dataset, Dataset, Dataset, Any]:
                 A tuple containing the datasets and the training result grid.
         """
-        train_with_data = tune.with_parameters(
+        import ray.train.torch
```
jarlsondre, line 1420: I might be wrong here, but couldn't you access this with `self.ray_train`?

jarlsondre:
> Looks god to me :)

Some might say divine, even

src/itwinai/torch/trainer.py
matbun: damn. You're right

src/itwinai/torch/distributed.py
annaelisalappe, line 1: Actually, I have one more general question, do you know generally the implications of moving imports into the classes? I wonder in cases such as ray train, where there would be multiple copies of the program running on each of the cores

jarlsondre: My general understanding is that the import is only valid in the scope that it's imported, meaning that if you import it in a class or function, then it is only valid in that class or function. That being said, each copy should run "the entire program" anyway, except of course the parts that are hidden behind if-statements (`rank == 0` etc.), so I wouldn't think that it's a problem. Not entirely sure of this, though.

annaelisalappe: Although it probably doesn't matter as the classes are each only instantiated once as well actually

src/itwinai/torch/trainer.py
matbun: yes, but then pylance would not resolve `TorchTrainer` unless I explicitly import `ray.train.torch`. So I could do `self.ray_train_torch = ray.train.torch` but since it's only used once I preferred to make it simple.

src/itwinai/torch/distributed.py
matbun: Considering that each copy is a separate process, it should not interfere with inter-process operations. AFAIU, Python will try to dynamically import packages only once per session, so even if you are hiding your import inside multiple a functions, it will actually get imported by Python only the first time and just reused the next times other functions import it.

src/itwinai/torch/trainer.py
jarlsondre: Yeah I see. Doesn't really matter anyway haha

jarlsondre (approved):
Looks divine to me

PR #286, opened 2025-01-10
Update itwinai user installation and remove itwinai installer

matbun:
# Summary

Remove Prov4ML dependency from pyproject.toml, as otherwise PyPI [complains](https://github.com/interTwin-eu/itwinai/actions/runs/12710939630/job/35433171220) when pushing a new release.

Updated user installation docs accordingly. 

---

**Related issue :**


README.md
```diff
@@ -180,17 +204,16 @@ directory, as you can very easily reach your disk quota otherwise. An example of
 complete command for installing as a developer on HPC with CUDA thus becomes:
 
 ```bash
-pip install -e ".[torch,dev,nvidia,tf]" \
+pip install -e ".[torch,dev,tf]" \
     --no-cache-dir \
     --extra-index-url https://download.pytorch.org/whl/cu121
 ```
 
-If you wanted to install this locally on macOS (i.e. without CUDA) with PyTorch, you
```
jarlsondre, line 188: One disadvantage of removing the macOS installation command is that the above command uses the CUDA wheel for PyTorch installation, which I believe can cause trouble when you try installing it on a computer without CUDA.

jarlsondre (approved):
Splendid!

PR #287, opened 2025-01-10
[DRAFT] add examples of scalability plots from virgo

jarlsondre:
Adding examples of the scalability plots to the `readthedocs` pages for the Virgo and EURAC use cases. Note that we do not add them to the READMEs, as this would mean that we would have to add them in two places, due to the incompatibility of importing markdown photos into rst.


annaelisalappe (approved):
Looks good to me

Issue #290, opened 2025-01-14
Test for building the docs


jarlsondre:
Currently, the docs are often forgotten (at least by myself) when working on features, making our `readthedocs` page contain errors etc. Therefore, it would be nice to check that the docs build properly whenever we create a pull request. Right now, when we do `make http` in `docs/`, it let's us build despite a bunch of warning. However, by adding `-W´to `SPHINXOPTS`, the build will fail if it encounters any errors. Then it should be possible to just add `cd docs/ && make clean && make http` to the GitHub Actions to check that the build works.

PR #291, opened 2025-01-15
Update building docs

jarlsondre:
# Summary
This PR contains three main changes:
- Add GitHub Action for building the docs
- Make the docs fail if it encounters any warnings (before it would build anyway)
- Add scalability plots from Virgo and EURAC using 1, 2 and 4 nodes in the documentation

Additionally, this solves a bug in the Virgo use case where the config was always overridden to be empty. 

**Related Issues**: #290


.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,45 @@
+name: Build Docs with Sphinx
+
+on:
+  push:
```
matbun, line 4: I would avoid this

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,45 @@
...
+        uses: actions/setup-python@v4
+        with:
+          python-version-file: .python-version
+
+      - name: Set up virtual environment
+        run: |
+          python -m venv .venv-docs
+          source .venv-docs/bin/activate
+          pip install --upgrade pip
+          pip install ".[torch,docs,linux]"
```
matbun, line 37: is linux extras still there?

.github/workflows/pytest.yml
```diff
@@ -34,5 +34,8 @@ jobs:
       # Default environment names are ".venv-pytorch" and ".venv-tf"
       - name: Run pytest for workflows
         shell: bash -l {0}
-        run: .venv-pytorch/bin/pytest -v ./tests/ -m "not hpc"
+        run: |
+          .venv-pytorch/bin/pip install prov4ml@git+https://github.com/matbun/ProvML@new-main &&
```
matbun, line 38: Please remove this from here. It should be in the generic install script. I am fixing this on another branch, tho.

docs/conf.py
```diff
@@ -54,10 +52,19 @@
...
 
 templates_path = ["_templates"]
-exclude_patterns = ["_build", "Thumbs.db", ".DS_Store"]
+exclude_patterns = [
+    "_build",
+    "Thumbs.db",
+    ".DS_Store",
+    "requirements.txt",
+    "README.md",
+    "uv-tutorial.md",
```
matbun, line 61: btw, it would be nice to have this tutorial in the docs webpage as well

docs/getting-started/plugins.rst
```diff
@@ -1,2 +1,5 @@
-.. include:: ../../tutorials/plugins/README.md
-   :parser: myst_parser.sphinx_
+Plugins
+=======
+
+.. .. include:: ../../tutorials/plugins/README.md
+..    :parser: myst_parser.sphinx_
```
matbun, line 5: I think this should be uncommented

docs/how-it-works/training/explain_ddp.rst
```diff
@@ -8,7 +8,7 @@ Even smaller DNNs can take days to train.
 Distributed Data Parallel (DDP) addresses these two issues, long training times and limited memory, by using multiple machines to host and train both model and data.
 
 Data parallelism is an easy way for a developer to vastly reduce training times.
-Rather than using single-node parallelism, Distributed Data Parallelism (DDP) scales to multiple machines. 
+Rather than using single-node parallelism, Distributed Data Parallelism scales to multiple machines.
```
matbun, line 11: DDP is mentioned below, so if you remove the acronym you should propagate the changes

use-cases/virgo/config.yaml
```diff
@@ -36,7 +36,7 @@ training_pipeline:
         init_args:
           config:
             generator: simple #unet
-            batch_size: 1
+            batch_size: ${batch_size}
```
matbun, line 39: batch size had been "forced" to 1  because the actual batch size was controlled chunk_size

docs/how-it-works/training/explain_ddp.rst
jarlsondre: "DDP" is actually defined in line 8 here, but line 11 should have used "DDP" instead of "Distributed Data Parallelism". I removed it because I saw it was defined twice, but forgot to actually use the acronym in line 11.

docs/getting-started/plugins.rst
jarlsondre: It might just be my branch or maybe I am just missing it, but the `tutorials/plugins/README.md` does not seem to exist

docs/conf.py
jarlsondre: Do you mean the `uv-tutorial`?

.github/workflows/pytest.yml
jarlsondre: I can remove it, but is it possible to approve the PR even with failing tests?

use-cases/virgo/config.yaml
jarlsondre: I am aware of that, but there is a `batch_size` variable at the top of the file, which is quite confusing. I guess I should look into whether it is used by the other pipelines or not, so that I potentially can remove the variable.

.github/workflows/build-docs.yml
jarlsondre: Nope, removed it now.

docs/getting-started/plugins.rst
jarlsondre: I actually saw now that it is a git submodule and I hadn't updated submodules. I added some stuff in the README, in the event that you had already cloned the repo before trying to build the docs.

docs/getting-started/plugins.rst
```diff
@@ -1,2 +1,5 @@
+Plugins
+=======
+
```
matbun, line 3: is this really needed?

.github/workflows/pytest.yml
matbun: Adding the following to generic torch install should fix the problem:

```python
# Install Prov4ML
if [[ "$(uname)" == "Darwin" ]]; then
  pip install --no-cache-dir  "prov4ml[apple]@git+https://github.com/matbun/ProvML@new-main"
else
  # Assuming Nvidia GPUs are available
  pip install --no-cache-dir  "prov4ml[nvidia]@git+https://github.com/matbun/ProvML@new-main"
fi
```

matbun (approved):

PR #296, opened 2025-01-17
Update Installation Docs

jarlsondre:
Currently, the installation docs for `itwinai` are rather messy. This PR cleans them up by porting them to Sphinx and rearranging the structure somewhat. 

**Relevant Issue(s)**: #297 

Tasks: 
- [x] Write the user installation
- [x] Write the developer installation
- [x] Check that the uv tutorial is up to date
- [x] Consider moving the uv tutorial to rst as well


docs/installation/developer_installation.rst
```diff
@@ -0,0 +1,103 @@
+Developer Installation
+======================
+**Author(s)**: Jarl Sondre Sæther (CERN)
```
matbun, line 3: I don't know if it makes sense to explicitly add the authors for the installation guide, considering that the current state is the result of different contributions

docs/installation/user_installation.rst
```diff
@@ -0,0 +1,70 @@
+User Installation (for Non-Developers)
+======================================
+**Author(s)**: Jarl Sondre Sæther (CERN)
```
matbun, line 3: I don't know if it makes sense to explicitly add the authors for the installation guide, considering that the current state is the result of different contributions

Issue #301, opened 2025-01-20
Unify and update the scalability report


jarlsondre:
Currently, there are three separate CLI functions that you have to call for the different parts of the scalability report: 
- `itwinai generate-scalability-plot`
- `itwinai generate-gpu-data-plots` 
- `itwinai generate-communication-plot`

We would like there to only be one... To be continued...

PR #302, opened 2025-01-21
Unify scalability report

jarlsondre:
# Summary
Updates the scalability report so that we only have a single CLI function instead of three distinct ones. Also cleans up the code a bit, makes the plots a bit prettier, and makes the plotting functions a bit more homogeneous. 

## The Updated Plots: 
absolute_epoch_time
relative_epoch_time_speedup

gpu_energy_plot
utilization_plot
computation_fraction_plot

**Related Issues**: #301


src/itwinai/cli.py
```diff
@@ -28,236 +28,92 @@
 
 
 @app.command()
-def generate_gpu_data_plots(
-    log_dir: str = "scalability-metrics/gpu-energy-data",
-    pattern: str = r".*\.csv$",
-    plot_dir: str = "plots/",
+def generate_scalability_report(
+    log_dir: str = "scalability-metrics",
+    plot_dir: str = "plots",
```
matbun, line 33: just noticed now. I think that you should use `Annotated` to document the arguments of CLI commands. This applies to all arguments in this function.

To test it run `itwinai generate-scalability-report --help`

src/itwinai/scalability_report/plot.py
matbun, line 1: add credits

src/itwinai/scalability_report/reports.py
matbun, line 1: add credits

src/itwinai/scalability_report/utils.py
matbun, line 1: add credits

matbun (commented):
Cool PR!

Do you want to add `itwinai.scalability_report` to the Python API section of the docs? if yes, you need to explicitly add it. I would add it to the list of modules imported by the sanity checkunder `itwinai.tests`

docs/python-api/modules.rst
```diff
@@ -16,3 +16,4 @@ itwinai
    itwinai.torch.modules
    itwinai.type
    itwinai.utils
+   itwinai.scalability_report.modules
```
matbun, line 19: modules should be sorted alphabetically

matbun (approved):

PR #305, opened 2025-01-22
Fix logger initialization

matbun:
# Summary

- improve the mechanism used to ensure that a logger is used after having been properly initialized
- make the logger wrapper/adapter for Lightning more robust by avoiding the corner case in which a logger is initialized twice
- add tests on initialization

---

**Related issue :** #283

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -55,6 +55,9 @@ jobs:
           VALIDATE_JSON_PRETTIER: false
           VALIDATE_CSS_PRETTIER: false
           VALIDATE_KUBERNETES_KUBECONFORM: false
+          # Both options below should be already covered by ruff
+          VALIDATE_PYTHON_ISORT: false
+          VALIDATE_PYTHON_FLAKE8: false
 
           # Only check new or edited files
           VALIDATE_ALL_CODEBASE: false
--- a/src/itwinai/loggers.py
+++ b/src/itwinai/loggers.py
@@ -368,6 +368,7 @@ class ConsoleLogger(Logger):
         self.worker_rank = rank
 
         if not self.should_log():
+            self.is_initialized = True
             return
 
         logging.info(f"Initializing {self.__class__.__name__} on rank {rank}")
@@ -526,6 +527,8 @@ class MLFlowLogger(Logger):
         self.run_name = run_name
         self.experiment_name = experiment_name
 
+        mfl_savedir.mkdir(parents=True, exist_ok=True)
+
         self.tracking_uri = (
             self.tracking_uri
             or os.environ.get("MLFLOW_TRACKING_URI")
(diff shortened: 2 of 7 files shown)


src/itwinai/loggers.py
```diff
@@ -672,18 +723,21 @@ def __init__(
...
-                used in distributed environments. Defaults to None.
+            rank (int): global rank of current process,
+                used in distributed environments. Defaults to 0.
         """
         self.worker_rank = rank
 
         if not self.should_log():
             return
 
+        logging.info(f"Initializing {self.__class__.__name__} on rank {rank}")
```
annaelisalappe, line 739: Have you thought about putting
```
 self.worker_rank = rank

        if not self.should_log():
            return

        logging.info(f"Initializing {self.__class__.__name__} on rank {rank}")
```
into the parent function? Then you can do `super(). create_logger_context(rank)` and you don't have to call it in every logger again.

tests/loggers/test_loggers.py
```diff
@@ -7,12 +7,51 @@
...
+    assert "has not been initialized" in str(exc_info.value)
+    with pytest.raises(RuntimeError) as exc_info:
+        itwinai_logger.destroy_logger_context()
+    assert "has not been initialized" in str(exc_info.value)
+
+    itwinai_logger.create_logger_context()
+
+    # Double initialization
+    caplog.clear()
+    with caplog.at_level(logging.WARNING):
```
annaelisalappe, line 45: This is cool :)

.github/workflows/lint.yml
```diff
@@ -55,6 +55,9 @@ jobs:
           VALIDATE_JSON_PRETTIER: false
           VALIDATE_CSS_PRETTIER: false
           VALIDATE_KUBERNETES_KUBECONFORM: false
+          # Both options below should be already covered by ruff
+          VALIDATE_PYTHON_ISORT: false
+          VALIDATE_PYTHON_FLAKE8: false
```
annaelisalappe, line 60: Is this from an old commit or did you change this?

matbun: I had to change it because isort + flake8 were conflicting with ruff on how to better lint some imports... idk if you ever noticed sth similar

src/itwinai/loggers.py
matbun: yes, good idea

matbun: I wouldn't know how to implement the "return" logic to skip logging when not supposed to.. If I return in super() I just go back to the caller

.github/workflows/lint.yml
annaelisalappe: Ah okay.. I had, but I changed these lines ages ago I though, that's why I asked if it was old - weird

matbun: ok. well, hopefully it'll stay this time :)

src/itwinai/loggers.py
annaelisalappe: Ah, I see. Well I suppose it's fine to leave it as it is as well 😌

PR #307, opened 2025-01-24
update eurac use case

iacopoff:
# Summary

This PR updates the Eurac use case pipeline to run with the refactored version of Hython.

Among other things, it adds the possibility to train a model with a probabilistic head layer, for uncertainty estimation.

The convlstm pipeline has been removed, and it will be reintegrated in the future.

---

**Related issue :**

### DIFF
--- a/use-cases/eurac/config.yaml
+++ b/use-cases/eurac/config.yaml
@@ -10,65 +10,195 @@
 # - Matteo Bunino <matteo.bunino@cern.ch> - CERN
 # --------------------------------------------------------------------------------------
 
+
 # General configuration
-rnn_surrogate_input: /p/scratch/intertwin/datasets/eurac/input/adg1km_eobs_preprocessed.zarr
-conv_surrogate_input: /p/scratch/intertwin/datasets/eurac/input/adg1km_eobs_original.zarr
-surrogate_model_output: /p/scratch/intertwin/datasets/eurac/model
-tmp_stats: /p/scratch/intertwin/datasets/eurac/stats
+experiment_name: "itwinai"
+experiment_run: "vwc_actevap"
 
-experiment: "drought use case lstm"
-run_name: "alps_test"
-epochs: 10
-random_seed: 1010
-lr: 0.001
-batch_size: 256
+work_dir: /p/scratch/intertwin/datasets/eurac/model
 
-target_names: ["vwc", "actevap"]
-strategy: ddp
+train_temporal_range: ["2016-01-01", "2018-12-31"]
+valid_temporal_range: ["2019-01-01", "2020-12-31"]
+test_temporal_range: ["2019-01-01", "2020-12-31"]
 
-rnn_training_pipeline:
+seed: 10 
+
+device: "cuda:0"
+
+# === Model ===
+
+model_file_name: model.pt
+
+model: cudalstm
+
+hidden_size: 136
+dropout: 0.4
+
+model_head_layer: regression # distr_normal
+model_head_activation: linear
+model_head_kwargs: null
+  
+# === Training ===
+
+strategy: ddp #null
+
+hython_trainer: rnntrainer
+
+loss_fn:
+  _target_: hython.losses.RMSELoss
+    
+metric_fn:
+  _target_: hython.metrics.MSEMetric
+
+optimizer: adam
+
+lr_scheduler:
+  mode: min
+  factor: 0.5
+  patience: 10
+
+seq_length: 120
+
+learning_rate: 0.001
+
+batch: 256
+
+epochs: 30
+
+gradient_clip: null
+
+target_weights: even # null, even, or dict
+
+# which steps are used in the computation of the loss function
+predict_steps: 0 # all # (prediction: 0 = t ), ( forecasts: 1 = t+1, 2 = t + 2)
+
+# > Donwsampling < 
+
+# spatial downsampling
+train_downsampler: null 
+  # _target_: hython.sampler.downsampler.RegularIntervalDownsampler
+  # intervals: [3,3]
+  # origin: [0,0]
+
+valid_downsampler: null 
+  # _target_: hython.sampler.downsampler.RegularIntervalDownsampler
+  # intervals: [3,3]
+  # origin: [1,1]
+
+test_downsampler: null
+  # _target_: hython.sampler.downsampler.RegularIntervalDownsampler
+  # intervals: [3,3]
+  # origin: [2,2]
+
+
+# temporal downsampling
+temporal_downsampling: True
+temporal_subset: [200, 150]
+
+
+# === Data ===
+
+dataset: Wflow1d
+
+data_source:
+  file:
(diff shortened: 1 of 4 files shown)


use-cases/eurac/config.yaml
```diff
@@ -10,65 +10,195 @@
...
+
+seq_length: 120
+
+learning_rate: 0.001
+
+batch: 256
+
+epochs: 30
+
+gradient_clip: null
```
jarlsondre, line 68: This is not super important, but I feel like the space between each line here makes this a little harder to read. I like the idea of organizing the file into different categories a lot, though, such as "Training".

use-cases/eurac/data.py
```diff
@@ -1,279 +1,56 @@
...
+        scaling_variant: str,
+        experiment_name: str,
+        experiment_run: str,
+        data_source: dict,
+        work_dir: str,
+        dynamic_inputs: list[str] = None,
+        static_inputs: list[str] = None,
+        target_variables: list[str] = None,
+        scaling_static_range: dict = None,
+        mask_variables: list[str] = None,
```
jarlsondre, line 28: I see two things that I think might be important to consider here: 

1) I would rather use `List` from the `typing` module than `list` here for the sake of conforming to normal Python convention. The same goes for `dict`; I would rather use `Dict`. 
2) These variables all have a value of `None`, but this is never indicated in the type hint. When you say that e.g. `dynamic_inputs` is of the type `list[str]`, the normal Python convention is that it is not possible for `dynamic_inputs` to have the value of `None`. However, the default argument is `None`, which would be a contradiction. To allow for it to be `None` as well, you would use the syntax `list[str] | None`.

use-cases/eurac/data.py
```diff
@@ -1,279 +1,56 @@
...
+        valid_temporal_range: list[str] = ["", ""],
+        train_downsampler: dict = None,
+        valid_downsampler: dict = None,
+        # == calibration ==
+        data_dynamic_inputs: str = None,
+        data_static_inputs: str = None,
+        data_target_variables: str = None,
+        data_target_mask: str = None,
+        min_sample_target: int = None,
+        seq_length: int = None,
```
jarlsondre, line 39: Same as above with the `None` typing.

use-cases/eurac/data.py
```diff
@@ -1,279 +1,56 @@
...
+        work_dir: str,
+        dynamic_inputs: list[str] = None,
+        static_inputs: list[str] = None,
+        target_variables: list[str] = None,
+        scaling_static_range: dict = None,
+        mask_variables: list[str] = None,
         train_temporal_range: list[str] = ["", ""],
-        test_temporal_range: list[str] = ["", ""],
-        name: str | None = None,
+        valid_temporal_range: list[str] = ["", ""],
```
jarlsondre, line 30: This might be fine in this specific application, but usually setting a list as a default value in the arguments is considered an anti-pattern in Python. This is because it is specifically bug prone, as the expression is only evaluated once. You can see an example of what I mean in this image: 
image
In this case, you can see that even though I call the function twice, the same dictionary remains. In your code the same will be the case. 

To avoid this, typically you would set the type to "List[str] | None = None" and then do a check in the code where you say
```python
if valid_temporal_range is None:
  valid_temporal_range = ["", ""]
```
This works because this is evaluated separately every time you call the function.

use-cases/eurac/trainer.py
```diff
@@ -104,64 +99,118 @@ def __init__(
...
+            )
 
-        target_weights = {
-            t: 1 / len(self.config.target_names) for t in self.config.target_names
-        }
-        self.loss_fn = RMSELoss(target_weight=target_weights)
-        self.metric_fn = MSEMetric()
+            surrogate.load_state_dict(
+                torch.load(
+                    f"{self.config.work_dir}/{self.config.model_head_dir}/{self.config.model_head_file}"
```
jarlsondre, line 151: In general, it is an advantage to separate these paths out into variables of their own. This is because it makes debugging easier and it allows it to be self-documented since you can give the variable an explanatory name. Additionally, I would use `pathlib`'s `Path` here, as that is considered best practice in modern Python.

use-cases/eurac/trainer.py
```diff
@@ -104,64 +99,118 @@ def __init__(
...
-    @profile_torch_trainer
+    
     @measure_gpu_utilization
     def train(self):
-        """Override version of hython to support distributed strategy."""
+        """Override train_val version of hython to support distributed strategy."""
+
         # Tracking epoch times for scaling test
         if self.strategy.is_main_worker:
             num_nodes = int(os.environ.get("SLURM_NNODES", "unk"))
```
jarlsondre, line 212: In this case you will run into trouble if "SLURM_NNODES" does not exist, as calling `int()` on the string "unk" will fail and thus crash the code.

use-cases/eurac/trainer.py
```diff
@@ -240,12 +270,22 @@ def train(self):
...
-            avg_metrics = pd.DataFrame(metric_history).mean().to_dict()
+            metric_history_ = {}
+            for period in metric_history:
+                for target in metric_history[period]:
+                    for key, value in target.items():
+                        l = []
+                        k = key.lower().split("metric")[0]
+                        n = period + "_" + k
+                        l.append(value)
+                        metric_history_[n] = l
```
jarlsondre, line 286: In general, you want to avoid single-letter variable names such as `l`, `k` and `n`, because it makes it very hard to understand what is going on. Generally, the cost of typing out an explanatory variable name is far far lower than the cost of trying to rewrite this code after forgetting what you were thinking when making this.

jarlsondre (changes requested):
From what I can tell, this looks mostly good. There are a couple of stylistic changes that could be considered, however. Personally, I would install `ruff` or `black` and utilize those to fix some of the stylistic stuff. Also, there is a flag you can add to `ruff` that makes it sort the imports as well, making the code extra clean. If you run `ruff` while inside the `itwinai` project, it will use our own configuration—which you can find in the `pyproject.toml`—that includes sorting the imports. Let me know if anything is unclear 😄

use-cases/eurac/config.yaml
iacopoff: @jarlsondre one way to organize better the config would be grouping sections into dict as:

```yaml
training:
    strategy: ddp
    hython_trainer: rnntrainer
    loss_fn:
      _target_: hython.losses.RMSELoss
    metric_fn:
      _target_: hython.metrics.MSEMetric
....
model:
   model_file_name: model.pt
   model: cudalstm
   hidden_size: 64
   dropout: 0.1
   model_head_layer: regression # distr_normal
   model_head_activation: linear
   model_head_kwargs: null
...
```
Another advantage is that in this way I can avoid prefixing keys with the section name to avoid potential conflicts with other keys. 

I will think about it, but would be interested in your opinion on that.

jarlsondre (approved):
Tested training with the new setup and it seems to work 👍

PR #321, opened 2025-02-05
CI pipeline v2

matbun:
# Summary

- spawn local interLink service to avoid reliance on a third-party deployment
- improve torch slim image, so that it can be run on HPC, while saving up some storage (sim is ~3.3GB, the torch image is ~12GBs)
- introduce torch skinny image, for lightweight itwinai container image with the bare minimum (~3.1GB) and quicker installation (no horovod and no deepspeed)
- introduce 2 CI pipelines: dev and release. The first is triggered when pushing to main, the latter when creating a new release. Release pipeline also runs tests on HPC. Both of them support conversion of Docker image to Singularity.
- Improved detection of distributed cluster
- fix problem with type hinting in itwinai.cli, as click does not support piped optionals

---

**Related issue :**


.github/actions/build-container/action.yml
```diff
@@ -10,30 +10,39 @@
...
+    description: "Avoid publishing Singularity image"
     required: false
+    default: "false"
   docker-token:
     description: "Docker secret auth token"
     required: true
-  kubeconfig-str:
-    description: "Kubeconfig secret"
+  interlink-values:
+    description: "Interlink VK configuration file (secret)"
```
jarlsondre, line 45: This is probably because I am not very familiar with interLink, but I'm not sure what "VK" means here. If it is an acronym, maybe it could be written out? This of course depends on what the convention is in this case

.github/workflows/container-ci.yml
```diff
@@ -27,21 +27,33 @@ jobs:
...
-            tag-template: "@{itwinai_version}-torch@{framework_version}-@{os_version}"
+          
+          # - base-img-name: "nvcr.io/nvidia/pytorch:24.05-py3"
+          #   dockerfile: "../env-files/torch/Dockerfile"
+          #   nickname: "torch"
+          #   tag-template: "@{itwinai_version}-torch@{framework_version}-@{os_version}"
+          #   skip-hpc: false
+          #   # Conversion to Singularity is skipped on GH actions as it would take too much space
+          #   skip-singularity: true
+          - base-img-name: "python:3.10-slim"
```
jarlsondre, line 38: Just curious: Why version 3.10 and not 3.11 here?

ci/examples.sh
```diff
@@ -13,51 +13,213 @@
 
 # Example of running dagger pipelines -- this script is mostly a scratchpad
 
+############## EXAMPLE OF WORKFLOWS ###############
+
+# IMPORTANT: when chaining steps that retrun Self, the chain is not really executed because
+# if lazy initialization if the steps. A workaround is appending the `logs` fuction at the
```
jarlsondre, line 19: "if" and "if" should probably be "of" and "of" here, no?

ci/src/main/hpc.py
```diff
@@ -0,0 +1,119 @@
...
+# Created by: Matteo Bunino
+#
+# Credit:
+# - Matteo Bunino <matteo.bunino@cern.ch> - CERN
+# --------------------------------------------------------------------------------------
+
+from typing import Annotated
+
+import dagger
+from dagger import Doc, dag, function, object_type
```
jarlsondre, line 13: Any particular reason for the mixed style of import here? What I mean is: Sometimes you seem to call `dagger.Container`, while other times you use `Doc` without the `dagger.` prefix. Why the distinction? Why not also import `Container` and any other tool from `dagger` as well? Is this due to convention?

ci/src/main/hpc.py
```diff
@@ -0,0 +1,119 @@
...
+        dag.container().from_("quay.io/singularity/docker2singularity")
+    )
+    docker: Annotated[
+        dagger.Container | None, Doc("Docker container to convert to Singularity")
+    ] = None
+
+    @function
+    def client(
+        self,
+    ) -> dagger.Container:
```
jarlsondre, line 30: The linter probably doesn't pick up on this due to the "," after `self`, but I believe this should be a single line.

ci/src/main/hpc.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @function
+    async def publish(
+        self,
+        password: Annotated[dagger.Secret, Doc("Password for registry")],
+        username: Annotated[dagger.Secret, Doc("Username for registry")],
+        uri: Annotated[
+            str,
+            Doc("Target URI for the image"),
+        ],
```
jarlsondre, line 58: I reckon the linter doesn't pick up on this due to the "," after `Doc(...)`, but this should probably be a single line.

ci/src/main/hpc.py
```diff
@@ -0,0 +1,119 @@
...
+            self.base_image.with_file("container.sif", self.convert())
+            .with_secret_variable(name="SINGULARITY_DOCKER_USERNAME", secret=username)
+            .with_secret_variable(name="SINGULARITY_DOCKER_PASSWORD", secret=password)
+            .with_exec(
+                [
+                    "singularity",
+                    "push",
+                    "container.sif",
+                    f"{uri}",
+                ]
```
jarlsondre, line 72: Same as above with the trailing comma: This should probably be one line.

ci/src/main/interlink.py
```diff
@@ -0,0 +1,344 @@
...
+
+import contextlib
+import dataclasses
+import time
+from datetime import datetime
+from typing import Annotated
+
+import dagger
+import yaml
+from dagger import Doc, dag, function, object_type
```
jarlsondre, line 18: Same question regarding the double `dagger` import as above. Is this because of convention?

ci/src/main/interlink.py
```diff
@@ -0,0 +1,344 @@
...
+                    "/values.yaml",
+                ]
+            )
+            # Wait for the VK to appear as a node
+            .with_exec(
+                [
+                    "sleep",
+                    f"{self.wait}",
+                ]
+            )
```
jarlsondre, line 112: Same as earlier comments with the trailing comma: This could probably be one line

ci/src/main/interlink.py
```diff
@@ -0,0 +1,344 @@
...
+        """Stop the k3s service on which interLink VK is running.
+
+        Returns:
+            dagger.Service: k3s service.
+        """
+        await self._service.stop()
+        self._service = None
+
+    @function
+    async def test_offloading(
```
jarlsondre, line 137: Shouldn't tests usually be put in a different file/directory? Or is this some other kind of test?

ci/src/main/interlink.py
```diff
@@ -0,0 +1,344 @@
...
+        self,
+        partition: Annotated[
+            str, Doc("HPC partition on which to test the offloading")
+        ] = "dev",
+    ) -> str:
+        """Test container offloading mechanism on remote HPC using interLink by runnign simple
```
jarlsondre, line 143: "runnign" -> "running"

ci/src/main/k8s.py
```diff
@@ -91,6 +85,7 @@ def create_pod_manifest(
     image_path: str = None,
     cmd_args: List[str] = None,
     resources: Dict = None,
+    target_node: str = None,
```
jarlsondre, line 88: Since these can be `None`, the type hint should say `<type> | None`, where `<type>` would be `str`, `Dict` etc.

ci/src/main/main.py
```diff
@@ -0,0 +1,587 @@
...
+        )
+
+    @function
+    def logs(self) -> str:
+        """Print the logs generted so far. Useful to terminate a chain of steps
+        returning Self, preventing lazy execution of it.
+        """
+        if not len(self._logs):
+            return "There are no logs to show"
+        return "\n\n".join(self._logs)
```
jarlsondre, line 101: This is a very small point, but it actually led me down into a PEP 8 rabbit hole. It turns out that `if not len(self._logs)` is not the Pythonic way of handling this, as seen here: https://peps.python.org/pep-0008/ 
image

Furthermore, however, in my voyage down the rabbit hole, I saw someone make a good point: "Make sure you actually _have_ to check the length of a list, because due to the strong idioms in Python, it usually won't be necessary". This is true, as `.join` will be empty if the length of the sequence is empty. This means that you could actually achieve the whole function with the following: 
```python
return "\n\n".join(self._logs) or "There are no logs to show"
``` 
I am not saying that you have to write it like this, but I thought this was pretty cool 😄

ci/src/main/main.py
```diff
@@ -0,0 +1,587 @@
...
+            ),
+        ] = None,
+    ) -> Self:
+        """Test container on remote HPC using interLink. Note that the container image must
+        already exist in some publicly accessible containers registry.
+        """
+
+        image = image or f"docker://{self.docker_registry}/{self.image}:{self.tag}"
+
+        if not image:
```
jarlsondre, line 266: I might be wrong here, but wouldn't the `not image` always evaluate as `False` here because of the definition above?

ci/src/main/main.py
```diff
@@ -0,0 +1,587 @@
...
+                name=pod_name, timeout=3000, poll_interval=30
+            )
+
+        if status not in ["Succeeded", "Completed"]:
+            message = (
+                f"Pod did not complete successfully! Status: {status}\n"
+                f"{"#"*100}\nJOB LOGS:\n\n{logs}"
+            )
+            print(message)
+            raise RuntimeError(message)
```
jarlsondre, line 364: If you at some point decide to handle this `RuntimeError` somewhere else, would you still want it to be printed into `stdout` as well? To me it seems natural to only print this if the exception stays uncaught, but maybe this is deliberate?

ci/src/main/main.py
```diff
@@ -0,0 +1,587 @@
...
+
+        if not skip_hpc:
+            if skip_singularity:
+                # The Docker image will be automatically converted to Singularity on HPC
+                # before running the tests
+                uri = f"docker://{self.singularity_registry}/itwinai-dev:{self.tag}"
+                await self.publish(usri=uri)
+            else:
+                assert username is not None, "Missing username for Singularity registry"
+                assert password is not None, "Missing password for Singularity registry"
```
jarlsondre, line 456: The same as the earlier comment regarding the assert. Maybe this could be put at the top of the function, unless you want to guarantee that the other stuff happens? I see that these asserts are repeated a bit further down too, so maybe by having them at the top you wouldn't have to have them twice?

docs/getting-started/containers.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Docker Containers (GHCR)
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Docker images are hosted on GitHub Container Registry (GHCR) under two repositories:
+
+- **Release Images**: ``ghcr.io/intertwin-eu/itwinai``
+- **Development Images**: ``ghcr.io/intertwin-eu/itwinai-dev``
+
+  - Includes images built from pushes to the ``main`` branch and during development
```
jarlsondre, line 37: Not super important but I am not quite sure what "during development" means here.

src/itwinai/distributed.py
```diff
@@ -39,6 +40,25 @@ class ClusterEnvironment(BaseModel):
     local_world_size: int = 1
 
 
+def detect_ray_cluster() -> bool:
```
jarlsondre, line 43: The one thing that has me a bit confused with the semantics of the name `detect_...` here is that it returns a bool, i.e. it checks if there exists a ray cluster or not. The function below—also starting with `detect_...`—however, actually retrieves the environment. Maybe this function should be called something like `ray_cluster_is_running` or something like that for consistency?

tests/torch/slurm.vega.sh
```diff
@@ -27,7 +29,7 @@
 #SBATCH --cpus-per-task=16
 #SBATCH --ntasks-per-node=1
 # SBATCH --mem-per-gpu=10G
-#SBATCH --exclusive
+# SBATCH --exclusive
```
jarlsondre, line 32: Why the space on this one in particular?

.github/actions/build-container/action.yml
matbun: it's a pretty standard acronym for virtual kubelet. I would leave it as is

.github/workflows/container-ci.yml
matbun: it's to keep consistency with the py version in all the other containers. The NGC container for torch 2.4 is still on py3.10, while the jlab container has to be on py3.10 for compatibility with jhub

ci/src/main/hpc.py
matbun: I find it easier to read because you immediately understand that sth is from dagger API (e.g., `dagger.File` is not some sort of standard or custom file type)

matbun: on the other hand, having `dagger.Doc` or `dagger.function` seems a bit too verbose.

matbun: eventually, I am following the dagger develpmnent style

ci/src/main/interlink.py
matbun: this is not a pytest test, just a function called like that ;)

ci/src/main/interlink.py
```diff
@@ -0,0 +1,344 @@
...
+    async def wait_pod(
+        self,
+        name: Annotated[str, Doc("pod name")],
+        timeout: Annotated[int, Doc("timeout in seconds")] = 300,
+        poll_interval: Annotated[int, Doc("how often to check the pod status")] = 5,
+    ) -> tuple[str, str]:
+        """Wait for pod termination (Succeeded, Failed) or timeout.
+        Returns last pod status and logs detected.
+        """
+        cnt = 0
```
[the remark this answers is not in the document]
matbun: easter egg :)

ci/src/main/main.py
matbun: cool! Nice find!

ci/src/main/main.py
matbun: Good catch!

ci/src/main/main.py
matbun: yes in this case is deliberate because of the way dagger manages stdout and stderr

ci/src/main/main.py
matbun: indeed, this can be moved up and the code can be simplified a bit. Good find

docs/getting-started/containers.rst
matbun: true, it can be safely removed

src/itwinai/distributed.py
matbun: yes, you're right, it's confusing with this name

tests/torch/slurm.vega.sh
matbun: it's just to "disable" it, without deleting the line.

Issue #312, opened 2025-02-03
Allow profiling to be toggled using a config flag instead of adding/removing a decorator


jarlsondre:
# Summary
If you wish to turn itwinai's profiling functionality on or off, you currently have to add or remove a decorator. This is quite cumbersome, as one shouldn't have to manually change the code just to toggle a configuration setting. Additionally, it is problematic as you also have to remove the import of the decorator in most cases—at least for the linter not to fail. 

**Potential Solution**: This can be solved by adding a field in the `TorchTrainer` that the decorator can use to decide whether or not to activate. This field can easily be passed to the `TorchTrainer` from a configuration, thus solving the problem.

PR #335, opened 2025-02-27
Toggle for Profilers

jarlsondre:
Having to add and remove the decorators can be quite cumbersome. Therefore, this PR adds the functionality of toggling them in the `TorchTrainer` constructor. 

**Related Issue**: #312


src/itwinai/loggers.py
```diff
@@ -1239,25 +1239,42 @@ def log(
...
             save_path = Path(save_path)
 
+        self.should_log = should_log
         self.save_path: Path = save_path
         self.strategy_name = strategy_name
         self.num_nodes = num_nodes
         self.data = {"epoch_id": [], "time": []}
 
+        if not self.should_log:
+            print("Warning: EpochTimeLogger has been disabled!")
```
matbun, line 1262: this may generate a large number of logs when running on multiple workers and logging only on rank 0. Maybe warn only the first time?

matbun (commented):
try to have your editor's formatter complying with the project's max line length

matbun (dismissed):

src/itwinai/torch/distributed.py
```diff
@@ -963,8 +963,8 @@ class NonDistributedStrategy(TorchDistributedStrategy):
 
     #: This strategy is not distributed.
     #: Defaults to False.
-    is_distributed: bool = True
     is_distributed: bool = False
+    is_initialized: bool = False
```
matbun, line 967: this should already be inherited from the parent `TorchDistributedStrategy`

jarlsondre: Removed

matbun (approved):

PR #338, opened 2025-03-04
Added calibration config and related updates to data and trainer

iacopoff:
This PR adds the config file and required trainer logic for running the calibration step of the eurac use case. 

There are now two configs, one for training the surrogate `config_training.yaml` and one for calibrating the surrogate `config_calibration.yaml`.

They should be run in sequence, first training and then calibration, as the calibration is loading and depends on the trained surrogate.

In theory they could be merged into one config.yaml and chained into a pipeline.

Noteworthy changes:
-  `data.py` instantiate a hython.Config class  and set attributes from the self.parameters of the itwinai's DataSplitter class. This was required because Hython accesses configs not only by keys (i.e. config['param1']) but as well by getattr (i.e. config.param1) .
- `trainer.py` updated the logic for logging model or model submodules locally or on mlflow tracking server


use-cases/eurac/data.py
```diff
@@ -1,62 +1,64 @@
...
         target_variables: List[str] | None = None,
         scaling_static_range: Dict | None = None,
         mask_variables: List[str] | None = None,
+        static_inputs_mask: List[str] | None = None,
+        head_model_inputs: List[str] | None = None,
         train_temporal_range: List[str] = None,
         valid_temporal_range: List[str] = None,
-        train_downsampler: Dict | None = None,
-        valid_downsampler: Dict | None = None,
+        cal_temporal_range: List[str] = None,
```
jarlsondre, line 35: Since you set the default value to `None`, you have to specify this in the type hints :)

use-cases/eurac/data.py
```diff
@@ -1,62 +1,64 @@
...
+        train_downsampler: AbstractDownSampler | None = None,
+        valid_downsampler: AbstractDownSampler | None = None,
+        cal_downsampler: dict = None,
```
jarlsondre, line 38: You should do `Dict | None` here

use-cases/eurac/trainer.py
```diff
@@ -290,17 +278,40 @@ def train(self):
...
+                        self.log(
+                            item=self.model.get_submodule(module_name),
+                            identifier=model_class_name,
+                            kind="model",
+                            registered_model_name=model_class_name,
+                        )
+                    else:
+                        self.model_api.log_model(
+                            module_name, self.model.get_submodule(module_name)
+                        )
```
jarlsondre, line 314: You can refactor this a bit by doing: 
```python
item = self.model if module_name == "model" else self.model.get_submodule(module_name)
if self.model_logger == "mlflow": 
  self.log(item=item, ...)
else:
 self.model_api.log_model(module_name, item)
```
This is kinda pseudocode but it would make your code a bit easier to read :)

jarlsondre (approved):
Looks good to me 😄

PR #346, opened 2025-03-11
Tutorial fixes #345

### LINKED ISSUE #345: Fix and update tutorial documentation
Many of the commands in the first tutorials are out of date. 

## Proposed fixes for [Tutorial page](https://itwinai.readthedocs.io/latest/tutorials/distrib-ml/)
- in 1.2: change to `--partition=develbooster` in `salloc` command (to work on juwels).
- in 1.2: for deepspeed use `torchrun` with `-s deepspeed` flag instead of `deepspeed` command.
- in 2.2: add `--partition=develbooster` flag to all `sbatch` commands

## Other
- small typo in [4.1.7.2](https://itwinai.readthedocs.io/latest/tutorials/distrib-ml/torch-tutorial-GAN.html#step-2-implement-distributed-training): itwinia
- NOTE tags (and others) are not rendered in sphinx -> think about making `.rst` files instead as proposed by @jarlsondre or find a plugin to render these tags from `.md` files.

lineick:
Admonitions in GitHub md syntax were not rendered properly in sphinx.
Fixed rendering issues of GitHub admonitions and fixed commands in tutorials/distributed-ml/.

# Summary
added admonition conversion script for readthedocs, which automatically runs when building the docs, fixed some commands in the tutorials, small typos and rst warnings.

---

**Related issue :**
Fixed Issue #345

### DIFF
--- a/docs/convert_admonitions.py
+++ b/docs/convert_admonitions.py
@@ -1,0 +1,148 @@
+# --------------------------------------------------------------------------------------
+# Part of the interTwin Project: https://www.intertwin.eu/
+#
+# Created by: Linus Eickhoff
+#
+# Credit:
+# - Linus Eickhoff <linus.maximilian.eickhoff@cern.ch> - CERN
+# --------------------------------------------------------------------------------------
+
+import argparse
+import re
+from pathlib import Path
+
+
+def convert_github_admonitions(content):
+    """
+    Converts GitHub-style admonitions to sphinx md syntax.
+
+    Args:
+        content (str): The markdown content containing GitHub-style admonitions
+
+    Returns:
+        str: The markdown content with GitHub admonitions converted to sphinx md syntax
+
+    Example:
+    > [!NOTE]
+    > This is a note
+
+    Becomes:
+    ```{note}
+    This is a note
+    ```
+    """
+    # Github markdown pattern
+    pattern = r">\s*\[!(NOTE|WARNING|TIP|IMPORTANT|CAUTION)\][ \t]*\n((?:>\s*[^\n]*\n)+)"
+
+    def replace_admonition(match):
+        # same name in lowercase for sphinx-compatible admonitions
+        admonition_type = match.group(1).lower()
+        content_lines = match.group(2)
+
+        cleaned_content = []
+        for line in content_lines.splitlines():
+            if line.strip():
+                # Preserve indentation after removing > prefix
+                cleaned_line = re.sub(r"^>\s?", "", line)
+                cleaned_content.append(cleaned_line)
+
+        return "```{" + f"{admonition_type}" + "}\n" + "\n".join(cleaned_content) + "\n```\n"
+
+    return re.sub(pattern, replace_admonition, content)
+
+
+def find_md_includes_in_rst(rst_content, rst_file_path):
+    """Find all .md file includes in an RST file and return their absolute paths.
+
+    Args:
+        rst_content (str): Content of the RST file to search for includes
+        rst_file_path (str): Path to the RST file, used for resolving relative paths
+
+    Returns:
+        list: List of absolute Path objects pointing to included .md files
+    """
+    pattern = r"\.\.\s+include::\s+(.+\.md)"
+    matches = re.finditer(pattern, rst_content)
+    md_files = []
+    rst_path = Path(rst_file_path)
+
+    for match in matches:
+        md_path = match.group(1)
+        # Handle relative paths properly using Path
+        abs_md_path = (rst_path.parent / md_path).resolve()
+        md_files.append(abs_md_path)
+
+    return md_files
+
+
+def process_rst_files(docs_dir):
+    """Process all RST files in the docs directory and convert referenced MD files.
+
+    Args:
+        docs_dir (str): Path to the docs directory containing RST files
+
+    Returns:
+        None: This function processes files in-place and prints status messages
+    """
+    docs_path = Path(docs_dir).resolve()
+    processed_files = set()  # To prevent processing the same file multiple times
+
+    for rst_file in docs_path.rglob("*.rst"):
+        if rst_file.is_symlink():  # Skip symlinks to prevent loops
+            print(f"Skipping symlink: {rst_file}")
+            continue
+
+        print(f"Processing RST file: {rst_file}")
+
+        try:
+            # Read RST content
+            rst_content = rst_file.read_text(encoding="utf-8")
+
+            # Find MD includes
+            md_files = find_md_includes_in_rst(rst_content, rst_file)
+
+            # Process each MD file
+            for md_file in md_files:
+                if md_file in processed_files:
+                    continue
+
+                processed_files.add(md_file)
+
+                if md_file.exists():
+                    print(f"Converting admonitions in: {md_file}")
+
+                    try:
+                        # assume uft8
+                        md_content = md_file.read_text(encoding="utf-8")
+
+                        # Convert admonitions
+                        converted_content = convert_github_admonitions(md_content)
+
+                        # Print to stdout where changes were made
+                        if converted_content != md_content:
+                            md_file.write_text(converted_content, encoding="utf-8")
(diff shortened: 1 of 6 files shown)


docs/convert_admonitions.py
```diff
@@ -0,0 +1,148 @@
...
+# --------------------------------------------------------------------------------------
+
+import argparse
+import re
+from pathlib import Path
+
+
+def convert_github_admonitions(content):
+    """
+    Converts GitHub-style admonitions to sphinx md syntax.
```
jarlsondre, line 17: Usually we don't have a newline after the triple quotes: https://peps.python.org/pep-0257/#handling-docstring-indentation

docs/convert_admonitions.py
```diff
@@ -0,0 +1,148 @@
...
+        str: The markdown content with GitHub admonitions converted to sphinx md syntax
+
+    Example:
+    > [!NOTE]
+    > This is a note
+
+    Becomes:
+    ```{note}
+    This is a note
+    ```
```
jarlsondre, line 32: Why do you not use this syntax instead? 

```
.. note::
```

It is the one we typically use when writing it ourselves. Of course, if it works with the above then no problem, mostly curious.

docs/convert_admonitions.py
```diff
@@ -0,0 +1,148 @@
...
+                # Preserve indentation after removing > prefix
+                cleaned_line = re.sub(r"^>\s?", "", line)
+                cleaned_content.append(cleaned_line)
+
+        return "```{" + f"{admonition_type}" + "}\n" + "\n".join(cleaned_content) + "\n```\n"
+
+    return re.sub(pattern, replace_admonition, content)
+
+
+def find_md_includes_in_rst(rst_content, rst_file_path):
```
jarlsondre, line 54: You should add programmatic type hints here, like mentioned above.

docs/convert_admonitions.py
```diff
@@ -0,0 +1,148 @@
...
+    for match in matches:
+        md_path = match.group(1)
+        # Handle relative paths properly using Path
+        abs_md_path = (rst_path.parent / md_path).resolve()
+        md_files.append(abs_md_path)
+
+    return md_files
+
+
+def process_rst_files(docs_dir):
```
jarlsondre, line 78: Type hints

docs/convert_admonitions.py
```diff
@@ -0,0 +1,148 @@
...
+                        converted_content = convert_github_admonitions(md_content)
+
+                        # Print to stdout where changes were made
+                        if converted_content != md_content:
+                            md_file.write_text(converted_content, encoding="utf-8")
+                            print(f"Updated: {md_file}")
+                    except Exception as e:
+                        print(f"Error processing {md_file}: {str(e)}")
+                else:
+                    print(f"Warning: Referenced MD file not found: {md_file}")
```
jarlsondre, line 128: A couple of things here: 
1. You can avoid a bit of indentation (and thus increase readability) by negating the if statement and doing print, followed by continue. 
2. Do you really need a try/except clause inside of an already existing try/except clause? At the end of your outer try/except clause, you can specify which exceptions you expect for each type and handle them separately there

jarlsondre: Ahh, I see why you have two try/except clauses: It's because of the double for loop. In which case, I would rather look for the culprit of the potential inner exception and exclude it with an if statement if possible. In other words, see if you can predict when something is going to go wrong. Is it the `write_text()` operation that you're worried about?

docs/convert_admonitions.py
lineick: the files are still md, they are just included via .rst files (and . . note : : is not markdown syntax, so it won't work)

docs/convert_admonitions.py
lineick: Yes, but I just grouped it together in the exceptions now, we don't need super specific error messages for this script.

tutorials/distributed-ml/torch-tutorial-0-basics/README.md
matbun, line 1: please update also the horvodrun command. I would explain that in general `horovodrun` is the suggested way to go according to the Horovod docs, but since it depends on mpirun it will not work when mpirun is not available, like on JSC.
When mpirun is not available, a valid fallback is srun. So please explain the users how to check if mpirun is there, and if not show the srun command.

Also, this would be a nice opportunity to include examples for Ray distributed strategies, or else please open an issue so that we don't forget.

docs/convert_admonitions.py
```diff
@@ -0,0 +1,145 @@
...
+    This is a note
+    ```
+    """
+    # Github markdown pattern
+    pattern = r">\s*\[!(NOTE|WARNING|TIP|IMPORTANT|CAUTION)\][ \t]*(.*(?:\n>.*)*)"
+
+    def replace_admonition(match):
+        # same name in lowercase for sphinx-compatible admonitions
+        admonition_type = match.group(1).lower()
+        content_lines = match.group(2)
```
jarlsondre, line 39: Calling `group(1)` and `group(2)` here assumes that there are two groups. Naturally, your regex above guarantees this, but if you ever change the regex then this could lead to an error here. Usually it's nice to have some explicit error handling for these kinds of things. In this case it might not be necessary, but certainly worth considering nonetheless.

docs/convert_admonitions.py
```diff
@@ -0,0 +1,145 @@
...
+    Returns:
+        list[Path]: List of absolute Path objects pointing to included .md files
+    """
+    pattern = r"\.\.\s+include::\s+(.+\.md)"
+    matches = re.finditer(pattern, rst_content)
+    md_files = []
+    rst_path = Path(rst_file_path)
+
+    for match in matches:
+        md_path = match.group(1)
```
jarlsondre, line 69: Same thing with the `group(1)` call as above. Probably not necessary to add a check here, but worth to think about anyway. I'll let you decide.

docs/convert_admonitions.py
```diff
@@ -0,0 +1,145 @@
...
+                if converted_content != md_content:
+                    md_file.write_text(converted_content, encoding="utf-8")
+                    print(f"Changed admonitions in: {md_file}")
+
+        except UnicodeDecodeError as e:
+            print(f"Error reading file {rst_file} or one of its included md files: {str(e)}")
+        except IOError as e:
+            print(f"Error accessing file {rst_file} or one of its included md files: {str(e)}")
+        except Exception as e:
+            print(f"Unexpected error processing {rst_file}: {str(e)}")
```
jarlsondre, line 127: One thing I was thinking about here: Are you sure you want to exit the inner loop if one of the `md_file` elements fails? I mean you're just printing the error anyway, but right now it will exit the inner for loop (potentially prematurely) and then continue with a new `rst_file` without necessarily having exhausted the previous one it was looking at. Just want to make sure that this is the behavior you intended.

src/itwinai/distributed.py
```diff
@@ -72,7 +73,8 @@ def detect_distributed_environment() -> ClusterEnvironment:
             local_world_size=os.getenv("LOCAL_WORLD_SIZE"),
             global_world_size=os.getenv("WORLD_SIZE"),
         )
-    elif os.getenv("OMPI_COMM_WORLD_SIZE") is not None:
+    # Fixes issue that OMPI_* env vars might be set despite using srun instead of mpirun
+    elif os.getenv("OMPI_COMM_WORLD_SIZE") > os.getenv("SLURM_NTASKS", 0):
```
jarlsondre, line 77: Can you guarantee that `OMPI_COMM_WORLD_SIZE` exists here? Because if it doesn't then `os.getenv()` will return `None`, which will cause an error here.

docs/convert_admonitions.py
lineick: As you say the regex guarantees this so putting explicit error handling for a case where this can't happen seems weird. If someone modifies the groups, he would have to add the exception.

matbun (approved):

PR #350, opened 2025-03-18
Added GAN validation metric: Frechet Inception Distance

lineick:
Added Frechet Inception Distance (FID) for proper GAN validation (e.g. for HPO etc.)

This branch was created from unified-trainers.


src/itwinai/torch/gan.py
```diff
@@ -604,46 +599,58 @@ def train_epoch(self):
...
+    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        """train step for GAN.
+
+        Args:
+            real_images (torch.Tensor): real images.
+            batch_idx (int): batch index.
+
+        Returns:
+            Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+            loss metrics of generator and discriminator and accuracy of the discriminator
```
jarlsondre, line 633: This should probably be indented one more, because now it looks like there are two return values

src/itwinai/torch/gan.py
```diff
@@ -665,139 +672,139 @@ def train_step(self, real_images, batch_idx):
...
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        """Validation step for GAN.
+
+        Args:
+            real_images (torch.Tensor): real images.
+            batch_idx (int): batch index.
+            fid (FrechetInceptionDistance): FID metric.
+
+        Returns:
+            Tuple[torch.Tensor, torch.Tensor]: accuracy of generator and discriminator.
```
jarlsondre, line 756: Same as above with listing them on different lines

src/itwinai/torch/gan.py
```diff
@@ -665,139 +672,139 @@ def train_step(self, real_images, batch_idx):
...
-            batch_idx=batch_idx,
-        )
+        accuracy_disc = accuracy / 2
+
+        # convert to 3 channel images for inceptionV3 model (which is used for FID) and
+        # use float64 for FID
+        real_images = real_images.repeat(1, 3, 1, 1).to(torch.float64)
+        fake_images = fake_images.repeat(1, 3, 1, 1).to(torch.float64)
+        fid.update(real_images, True)
+        fid.update(fake_images, False)
```
jarlsondre, line 787: It would be better to use keyword arguments here, so that the meaning of `True` and `False` become obvious, instead of having to look it up.

jarlsondre (approved):

PR #354, opened 2025-03-26
Add CaloINN use-case (dalatoading and training)

VeraMay:
# Summary

CaloINN use-case (dalatoading and training steps, no inference)
---

**Related issue :**


use-cases/caloinn/config.yaml
```diff
@@ -0,0 +1,79 @@
...
+      dataset_type: ${dataset_type}
+      eps: 1.0e-10
+      u0up_cut: 3.5 
+      u0low_cut: 0.0 
+      dep_cut: 600
+      width_noise: 5.0e-6 
+      fixed_noise: False
+      noise: True
+      validation_proportion: 0.01
+      #rnd_seed: 42
```
jarlsondre, line 30: I see that you also have `random_seed` further down (also commented out). Should both of these be here?

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+
+        self.data_train_url = type_to_link_train[self.dataset_type]
+        self.data_test_url  = type_to_link_test[self.dataset_type]  
+
+    @monitor_exec
+    def execute(self):
+        # Download data
+        if not os.path.exists(self.data_path):
+            if self.data_train_url is None:
+                print("WARNING! Train data URL is None. " "Skipping train dataset downloading")
```
jarlsondre, line 45: Do you need the space between the strings here? I'm referring to `print("..." "...")`.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+            if self.data_test_url is None:
+                print("WARNING! Test data URL is None. " "Skipping test dataset downloading")
```
jarlsondre, line 47: Same as above with the space between the strings.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+
+        def download_rename(url_list, prefix):
```
jarlsondre, line 49: I am a little confused by the semantics of "rename" here: What makes it a new file name? Is there an existing file name somewhere?

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+            for url in url_list:
+                response = requests.get(url, stream=True)
+                if response.status_code == 200:
+                    new_filename = f"{url.split('/')[-1].split('.')[0]}_{prefix}.hdf5"
+                    with open(os.path.join(self.data_path, new_filename), 'wb') as file:
+                    # Write the content of the response to the file
+                        for chunk in response.iter_content(chunk_size=8192):
+                            file.write(chunk)
+                    print(f"File downloaded and saved as: {new_filename}")
```
jarlsondre, line 58: It might be better to include information about exactly which file was downloaded here. I suppose some of it might be contained in the `new_filename`, but it could be beneficial to explicitly write the URL in the message for easier debugging.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+import torch
+from src.XMLHandler import XMLHandler
+from itwinai.components import DataGetter, monitor_exec, DataSplitter
+
+
+class CaloChallengeDownloader(DataGetter):
+    def __init__(
+        self,
+        data_path: Optional[str] = "./calochallenge_data/",
+        dataset_type: Optional[str] = "dataset_1_photons",
```
jarlsondre, line 16: Usually we prefer using `str | None` instead of `Optional[str]` because semantically the word "optional" seems to imply that you don't need to pass the parameter—i.e. that it has a default value—but `Optional` from typing literally means "could be `None`" and that's how your type checker is going to interpret it. This is purely a preference, though.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
+import os
+from typing import Optional, Tuple
+import h5py
+import numpy as np
+import requests
+from torch.utils.data import Dataset, random_split
+import torch
+from src.XMLHandler import XMLHandler
+from itwinai.components import DataGetter, monitor_exec, DataSplitter
```
jarlsondre, line 9: It's better if you sort these. You can use `isort` or `ruff` to do this. If you use `ruff` while in the `itwinai` project it will automatically sort imports for you. If you're not inside `itwinai` then you have to add some options to `ruff`.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+    ) -> None:
+        self.save_parameters(**self.locals2params(locals()))
+        super().__init__()
+        self.data_path = data_path
+        self.dataset_type = dataset_type
+
+        type_to_link_train = {"dataset_1_photons" : ["https://zenodo.org/records/8099322/files/dataset_1_photons_1.hdf5"],
+                              "dataset_1_pions" : ["https://zenodo.org/records/8099322/files/dataset_1_pions_1.hdf5"],
+                              "dataset_2" : ["https://zenodo.org/records/6366271/files/dataset_2_1.hdf5"],
+                              "dataset_3" : ["https://zenodo.org/records/6366324/files/dataset_3_1.hdf5", "https://zenodo.org/records/6366324/files/dataset_3_2.hdf5"]}
```
jarlsondre, line 26: Generally, when we have multiple lines inside a list or dictionary, we always have a newline after the opening bracket and a newline before the closing one. This means that instead of doing this: 
```python
my_list = [item1,
                 item2]
```
we would do this:
```python
my_list = [
    item1,
    item2
]
```
Obviously in the example above it would be silly to use multiple lines, but in general it helps readability because it moves everything much further to the left. This can be easily fixed if you use the `black` formatter, though. Worth looking into, can be installed via `pip install black` (I believe)

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+        
+        type_to_link_test  = {"dataset_1_photons" : ["https://zenodo.org/records/8099322/files/dataset_1_photons_2.hdf5"],
+                              "dataset_1_pions" : ["https://zenodo.org/records/8099322/files/dataset_1_pions_2.hdf5"],
+                              "dataset_2" : ["https://zenodo.org/records/6366271/files/dataset_2_2.hdf5"],
+                              "dataset_3" : ["https://zenodo.org/records/6366324/files/dataset_3_3.hdf5", "https://zenodo.org/records/6366324/files/dataset_3_4.hdf5"]}
+
+        if dataset_type not in type_to_link_train.keys():
+            print("WARNING! Dataset type is invalid " "Loading dataset 1 photon")
```
jarlsondre, line 34: There is a space between the strings, but you could make this a single string (as mentioned earlier)

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+from torch.utils.data import Dataset, random_split
+import torch
+from src.XMLHandler import XMLHandler
+from itwinai.components import DataGetter, monitor_exec, DataSplitter
+
+
+class CaloChallengeDownloader(DataGetter):
+    def __init__(
+        self,
+        data_path: Optional[str] = "./calochallenge_data/",
```
jarlsondre, line 15: Since this is a path, it would be nice to use `Path` from `pathlib` as a type instead. The best thing to do would be to allow for both by doing `str | Path` and then converting to `Path` when you store it as `self.` if it is of type `str`. Something like this: 
```python
def __init__(..., data_path: Path | str | None = ...):
...
if isinstance(data_path, str): 
    data_path = Path(data_path)
self.data_path = data_path
```

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+            print("WARNING! Dataset type is invalid " "Loading dataset 1 photon")
+            self.dataset_type = "dataset_1_photons"
+
+        self.data_train_url = type_to_link_train[self.dataset_type]
+        self.data_test_url  = type_to_link_test[self.dataset_type]  
+
+    @monitor_exec
+    def execute(self):
+        # Download data
+        if not os.path.exists(self.data_path):
```
jarlsondre, line 43: If you make `self.data_path` into a `Path` object (as mentioned earlier) then you can just do `self.data_path.exists()` here and avoid using `os` altogether.

use-cases/caloinn/data.py
jarlsondre: Also you should add type hints to the function. What is the expected type of `url_list` and `prefix`, and what does `download_rename` return?

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+            if self.data_train_url is None:
+                print("WARNING! Train data URL is None. " "Skipping train dataset downloading")
+            if self.data_test_url is None:
+                print("WARNING! Test data URL is None. " "Skipping test dataset downloading")
+
+        def download_rename(url_list, prefix):
+            for url in url_list:
+                response = requests.get(url, stream=True)
+                if response.status_code == 200:
```
jarlsondre, line 52: In this case you could negate the if statement and handle the `else` first, and then use the `continue` keyword to avoid a level of indentation. An example: 
```python
if response.status_code != 200: 
  print("failed ...")
  continue

new_file_name = ...
```

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+                    new_filename = f"{url.split('/')[-1].split('.')[0]}_{prefix}.hdf5"
+                    with open(os.path.join(self.data_path, new_filename), 'wb') as file:
```
jarlsondre, line 54: If `self.data_path` is of type `Path` (as mentioned earlier) then you can do `self.data_path / new_filename` here instead of `os.path.join`, which makes it a bit simpler.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+        print(f"Downloading test dataset to {self.data_path}")
+        download_rename(self.data_test_url, "test")
+
+
+class CalochallengeDataset(Dataset):
+    def __init__(
+        self, 
+        data_path: str, 
+        dataset_type: str, 
+        dataset_trainortest: str,
```
jarlsondre, line 73: It is not immediately obvious to me what `dataset_trainortest` means here. It seems—based on what's below—that it is an indicator for whether you should use the `train` or `test` partition. However, since this is binary you might be better off just doing `train: bool` or something along those lines. If you wish to keep the name, then I would at least add underscores to make it read a little better, i.e. `dataset_train_or_test`.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+        else:
+            self.dataset_trainortest = dataset_trainortest
+
+        torch.set_default_dtype(torch.float32)
+
+        # Create a XML_handler to extract the layer boundaries. (Geometric setup is stored in the XML file)
+        dataset_to_particle_type = {"dataset_1_photons" : "photon",
+                                    "dataset_1_pions" : "pion",
+                                    "dataset_2" : "electron",
+                                    "dataset_3" : "electron"}
```
jarlsondre, line 94: Same stuff as mentioned earlier with the newline after brackets.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+        eps: float = 1e-10,
+        u0up_cut: float = 7.0, 
+        u0low_cut: float = 0.0, 
+        dep_cut: float = 1e10,
+        width_noise: float = 1e-7, 
+        fixed_noise: bool = False,
+        noise: bool = False
+        ) -> None:
+        if dataset_trainortest not in ["train", "test"]:
+            print("WARNING! Subset type (train ir test) is not defined" "Loading as train")
```
jarlsondre, line 83: There is a space between the strings here.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+            self.dataset_trainortest = dataset_trainortest
+
+        torch.set_default_dtype(torch.float32)
+
+        # Create a XML_handler to extract the layer boundaries. (Geometric setup is stored in the XML file)
+        dataset_to_particle_type = {"dataset_1_photons" : "photon",
+                                    "dataset_1_pions" : "pion",
+                                    "dataset_2" : "electron",
+                                    "dataset_3" : "electron"}
+        if dataset_type not in dataset_to_particle_type.keys():
```
jarlsondre, line 95: Since `dataset_to_particle_type` is a dict, you don't really need to call `keys()` at the end here.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+            print("WARNING! Dataset type is invalid " "Loading dataset 1 photon")
```
jarlsondre, line 96: Space between strings.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+            self.dataset_type = "dataset_1_photons"
+        else:
+            self.dataset_type = dataset_type
+
+        xml_handler = XMLHandler(particle_name=dataset_to_particle_type[self.dataset_type], 
+                                 filename="./calochallenge_binning/binning_" + self.dataset_type + ".xml")
```
jarlsondre, line 102: You should have a newline after the opening parenthesis here. Also you should consider making the `filename` here into a separate variable.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+            self.noise_distribution = torch.distributions.Uniform(torch.tensor(0.), torch.tensor(1.))
+        
+
+    def __len__(self):
+        return self.x.shape[0]
+
+    def __getitem__(self, idx):
+        data = self.x[idx]
+        if not self.fixed_noise and self.noise:
+            noise = self.noise_distribution.sample(data.shape)*self.width_noise
```
jarlsondre, line 129: Generally you would want to have a space before and after operators such as "*". Formatters would probably fix this though.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+        self.noise = noise
+        self.width_noise = width_noise
+        self.fixed_noise = fixed_noise
+        
+        self.data = {}
+
+        self._load_data()
+        self._preprocess_data(eps, u0up_cut, u0low_cut, dep_cut)
+        if self.fixed_noise and self.noise:
+            noise = self.noise_distribution.sample(self.x.shape)*self.width_noise
```
jarlsondre, line 116: I can't seem to find where you define `self.x` in this class. Are you sure you've defined it before this part of the code?

Also, there should be a space before and after the "*"

jarlsondre: It would also be nice with a name that's a bit more descriptive.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+        if not self.fixed_noise and self.noise:
+            noise = self.noise_distribution.sample(data.shape)*self.width_noise
+            data = (torch.tensor(data) + noise.reshape(data.shape).clone()).clone()
+        cond = self.cond[idx]
+        return data, cond
+
+    def _load_data(self) -> None:
+        print("Searching in :", self.data_path)
+        files = [os.path.join(self.data_path, filename) for filename in os.listdir(self.data_path) if self.dataset_type in filename and self.dataset_trainortest in filename] 
+        print("Found {} files. ".format(len(files)))
```
jarlsondre, line 137: You might be better off using an f-string here for clarity. Not super important though

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+        if len(files) == 0:
+            raise RuntimeError(f"No H5 files found at '{self.data_path}'!")
```
jarlsondre, line 139: Usually you want to avoid using `RuntimeError` because semantically that just means "the code crashed" which doesn't really say all that much by itself. You would be better off using `FileNotFoundError` here.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+
+        for file in files:
+            with h5py.File(file, "r") as data_file:
+                if "energy" not in self.data:
+                    self.data["energy"] = data_file["incident_energies"][:] / 1.e3
+                else:
+                    self.data["energy"] = np.concatenate((self.data["energy"], data_file["incident_energies"][:] / 1.e3))
+
+                for layer_index, (layer_start, layer_end) in enumerate(zip(self.layer_boundaries[:-1], self.layer_boundaries[1:])):
```
jarlsondre, line 148: This line is a little too long, see if you can make it shorter.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+                    self.data[f"layer_{layer_index}"] = data_file["showers"][..., layer_start:layer_end] / 1.e3
+
+        print(self.data[f"layer_{layer_index}"].shape)
```
jarlsondre, line 151: This seems like a print without context. You should probably add some context to the print statement to explain what we are looking at or consider removing the print altogether.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+
+    def _preprocess_data(self, eps, u0up_cut, u0low_cut, dep_cut) -> None:
```
jarlsondre, line 153: You should add some type hints to this function.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+
```
jarlsondre, line 154: Remove the empty line before the docstring.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+        """Transforms the dict 'data' into the ndarray 'x'. Furthermore, the events
+        are masked and the extra dims are appended to the incident energies"""
```
jarlsondre, line 156: Generally, when using a multi-line docstring, you want to have the last `""" ` on a new line. The starting `"""` is correct. Read PEP 257 for more information: https://peps.python.org/pep-0257/#multi-line-docstrings

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+        extra_dims = extra_dims[binary_mask]
+        print("final shape of dataset: ", x.shape)
+
+        x = self.normalize_layers(x, c, eps)
+        x = np.concatenate((x, extra_dims), axis=1)
+
+        self.x = x
+        self.cond = c
+
+    def get_energy_and_sorted_layers(self):
```
jarlsondre, line 186: Add a type hint for the return type

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+        """returns the energy and the sorted layers from the data dict"""
+        
+        # Get the incident energies
+        energy = self.data["energy"]
+
+        # Get the number of layers layers from the keys of the data array
+        number_of_layers = len(self.data)-1
+        
+        # Create a container for the layers
+        layers = []
```
jarlsondre, line 196: These comments might be a bit redundant, as the variable names here are pretty good and thus self-documenting.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+
+        # Append the layers such that they are sorted.
+        for layer_index in range(number_of_layers):
+            layer = f"layer_{layer_index}"
+            
+            layers.append(self.data[layer])
+        
+        return energy, np.concatenate(layers, axis=1)
+    
+    def get_energy_dims(self, x, c, eps):
```
jarlsondre, line 206: Add type hints

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+        """Appends the extra dimensions and the layer energies to the conditions
+        The layer energies will always be the last #layers entries, the extra dims will
+        be the #layers entries directly after the first entry - the incident energy.
+        Inbetween additional features might be appended as further conditions"""
```
jarlsondre, line 210: Final `"""` on new line.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+        for layer_start, layer_end in zip(self.layer_boundaries[:-1], self.layer_boundaries[1:]):
+            
+            # Compute total energy of current layer
+            layer_energy = np.sum(x[..., layer_start:layer_end], axis=1, keepdims=True)
+            
+            # Normalize current layer
+            x[..., layer_start:layer_end] = x[..., layer_start:layer_end] / (layer_energy + eps)
+            
+            # Store its energy for later
+            layer_energies.append(layer_energy)
```
jarlsondre, line 226: This comment might be a bit redundant as the code speaks for itself

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+            
+        layer_energies_np = np.array(layer_energies).T[0]
+            
+        # Compute the generalized extra dimensions
+        extra_dims = [np.sum(layer_energies_np, axis=1, keepdims=True) / c]
+
+        for layer_index in range(len(self.layer_boundaries)-2):
+            extra_dim = layer_energies_np[..., [layer_index]] / (np.sum(layer_energies_np[..., layer_index:], axis=1, keepdims=True) + eps)
```
jarlsondre, line 234: This line is a bit too long

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+            extra_dims.append(extra_dim)
+            
+        # Collect all the conditions
+        extra_dims = np.concatenate(extra_dims, axis=1)
+
+        return c, extra_dims
+
+    def normalize_layers(self, x, c, eps=1.e-10):
```
jarlsondre, line 242: You should add type hints to this function

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,313 @@
...
+                                                  eps=eps,
+                                                  u0up_cut=u0up_cut,
+                                                  u0low_cut=u0low_cut,
+                                                  dep_cut=dep_cut,
+                                                  noise=False
+                                                  )
+    @monitor_exec
+    def execute(self) -> Tuple:
+        # Split train file into train and validation sets
+        [train_dataset, validation_dataset] = random_split(self.train_dataset,
```
jarlsondre, line 309: You probably don't need the brackets around `train_dataset` and `validation_dataset`

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+    def __init__(self, num_layers, size_in, size_out, internal_size=None, dropout=0.0,
+                 layer_class=nn.Linear, layer_args={}, layer_norm=None, layer_act="nn.ReLU"):
+        """
+            Initializes subnet class.
+
+            Parameters:
+            size_in: input size of the subnet
+            size: output size of the subnet
+            internal_size: hidden size of the subnet. If None, set to 2*size
+            dropout: dropout chance of the subnet
```
jarlsondre, line 29: `size_in`, `size` etc. should probably be indented one level more than "Parameters.". Alternatively, you could follow the Google Python Style Guide: https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+from src.vblinear import VBLinear
+
+from src.XMLHandler import XMLHandler
+import h5py
+
+from src.spline_blocks import CubicSplineBlock, RationalQuadraticSplineBlock
+from src.made import MADE
+
+class Subnet(nn.Module):
+    """ This class constructs a subnet for the coupling blocks """
```
jarlsondre, line 18: Remove the space before "This"

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+        self.layer_list = []
+        for n in range(num_layers - 1):
+            if isinstance(internal_size, list):
+                input_dim, output_dim = internal_size[n], internal_size[n+1]
+            else:
+                input_dim, output_dim = internal_size, internal_size
+            if n == 0:
+                input_dim = size_in
+
+            self.layer_list.append(layer_class[n](input_dim, output_dim, **(layer_args[n])))
```
jarlsondre, line 45: The contents of `append()` here should probably be split into a variable (or two) as it is quite hard to understand what is going on the way it is currently written.

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+        final_layer_name = str(len(self.layers) - 1)
+        for name, param in self.layers.named_parameters():
+            if name[0] == final_layer_name and "logsig2_w" not in name:
+                param.data.zero_()
+
+    def forward(self, x):
+        return self.layers(x)
+
+class NormTransformation(fm.InvertibleModule):
+    def __init__(self, dims_in, dims_c=None, log_cond=False):
```
jarlsondre, line 69: Make sure to add type hints here, especially for the variable with a default to `None`

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+        c, = c
+        if self.log_cond:
+            c = torch.exp(c)
+        if rev:
+            z = x/c
+            jac = -torch.log(c)
+        else:
+            z = x*c
+            jac = torch.log(c)
+        return (z, ), torch.tensor([0.], device=x.device)
```
jarlsondre, line 84: You should add type hints and give the variables more descriptive names here

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def output_dims(self, input_dims):
+        return input_dims
```
jarlsondre, line 87: What is the point of this function?

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+    
+class LogTransformation(fm.InvertibleModule):
+    def __init__(self, dims_in, dims_c=None, alpha = 0., alpha_logit=0.):
+        super().__init__(dims_in, dims_c)
+        self.alpha = alpha
+        self.alpha_logit = alpha_logit
```
jarlsondre, line 93: Type hints

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def forward(self, x, c=None, rev=False, jac=True):
+        x, = x
+        if rev:
+            z = torch.exp(x) - self.alpha
+            jac = torch.sum( x, dim=1)
+        else:
+            z = torch.log(x + self.alpha)
+            jac = - torch.sum( z, dim=1)
+        return (z, ), torch.tensor([0.], device=x.device) # jac
```
jarlsondre, line 103: Type hints

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def output_dims(self, input_dims):
+        return input_dims
```
jarlsondre, line 106: What is the point of this function?

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+
+class LogUniform(dist.TransformedDistribution):
+    def __init__(self, lb, ub):
+        super(LogUniform, self).__init__(dist.Uniform(torch.log(lb), torch.log(ub)),
+                                            dist.ExpTransform())
+
+class CINN(nn.Module):
+    """ cINN model """
+
+    def __init__(self, data_dim, config):
```
jarlsondre, line 116: Type hints

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+        """ Initializes model class.
+
+        Parameters:
+        params: Dict containing the network and training parameter
+        data: Training data to initialize the norm layer
+        cond: Conditions to the training data
```
jarlsondre, line 122: Indentation?

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+class LogUniform(dist.TransformedDistribution):
+    def __init__(self, lb, ub):
+        super(LogUniform, self).__init__(dist.Uniform(torch.log(lb), torch.log(ub)),
+                                            dist.ExpTransform())
+
+class CINN(nn.Module):
+    """ cINN model """
+
+    def __init__(self, data_dim, config):
+        """ Initializes model class.
```
jarlsondre, line 117: Remove space before "Initializes"

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+        params: Dict containing the network and training parameter
+        data: Training data to initialize the norm layer
+        cond: Conditions to the training data
+        """
+        super(CINN, self).__init__()
+    
+        self.num_dim = data_dim
+        self.config = config
+
+        self.norm_m = None
```
jarlsondre, line 129: What does the "m" in `norm_m` refer to? Might just be that I'm not well-versed in the domain though

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+        self.use_norm = self.config.use_norm and not self.config.use_extra_dim
+        self.pre_subnet = None
+        self.bayesian = self.config.bayesian
+        self.pre_subnet = None
```
jarlsondre, line 133: This is already defined above. Also, since you're setting it to None you should add a type hint so that we know what to expect it to be later.

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+
+        self.sub_layers = self.config.sub_layers
+
+        if self.config.bayesian:
+            self.bayesian_layers = []
+
+        self.define_model_architecture(self.num_dim)
+        print(self.model)
+
+    def initialize_normalization(self, data, cond):
```
jarlsondre, line 143: Type hints

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+        data = torch.clone(data)
+        if self.use_norm:
+            data /= cond
+        data = torch.log(data + self.config.alpha)
+        mean = torch.mean(data, dim=0)
+        std = torch.std(data, dim=0)
+        self.norm_m = torch.diag(1 / std)
+        self.norm_b = - mean/std
+
+        print('num samples out of bounds:', torch.count_nonzero(torch.max(torch.abs(data), dim=1)[0] > self.params.get("bounds_init", 10)).item())
```
jarlsondre, line 154: This line is a bit too long, you should create a separate variable for the oob calculation

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+            raise ValueError(f"Unknown Coupling block type {self.config.coupling_type}")
+        
+        return CouplingBlock, block_kwargs
+
+    def get_constructor_func(self):
+        """ Returns a function that constructs a subnetwork with the given parameters """
+        if self.sub_layers:
+            layer_class = self.get_layer_class()
+            layer_args = self.get_layer_args()
+        else:
```
jarlsondre, line 261: If you add a `return` at the end of the if statement you can remove the `else` and all the indentation that comes with it, for more readable code.

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+                    layer_class = layer_class,
+                    layer_args = layer_args,
+                    layer_norm = self.config.layer_norm or None,
+                    layer_act = self.config.layer_act or nn.ReLU,
+                    )
+            if self.config.bayesian:
+                self.config.bayesian_layers.extend(
+                    layer for layer in subnet.layer_list if isinstance(layer, VBLinear))
+            return subnet
+        return func
```
jarlsondre, line 289: You might be able to replicate this using `functool`'s `partial`, as that's generally the recommended way of doing "currying" like you are here

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def get_layer_class(self):
+        lays = []
+        for n in range(len(self.sub_layers)):
+            if self.sub_layers[n] == 'vblinear':
+                lays.append(VBLinear)
+            if self.sub_layers[n] == 'linear':
+                lays.append(nn.Linear)
+        return lays
```
jarlsondre, line 298: Are these the only two possible layers in `self.sub_layers`? If this is the case, then there are a couple of options to make this a bit simpler. First option would be do just do: 
```python
lays = [VBLinear for elem in sub_layers if elem == "vblinear" else nn.Linear]
```
to get it in a single line. Another thing to do would be add a fail safe, i.e. checking that if you don't go into either of the if branches then throw an error.

Whether any of these approaches are applicable depends entirely on the assumption that "vblinear" and "linear"  are the only two options, though.

use-cases/caloinn/src/XMLHandler.py
```diff
@@ -0,0 +1,149 @@
...
+        return self.eta_all_layers, self.phi_all_layers
+
+    def GetRelevantLayers(self):
+        return self.relevantlayers
+
+    def GetLayersWithBinningInAlpha(self):
+        return self.layerWithBinningInAlpha
+
+    def GetEtaRegion(self):
+        return self.etaRegion
```
jarlsondre, line 149: Generally you'd want to avoid camelCase in Python as the standard is snake_case. Additionally, upper-case camelCase is usually reserved for classes. 

Disregarding the style, there is usually no reason to have getters in Python unless you have some specific logic you want to add. In this case you're just returning the variables directly, so you could've just omitted the getters altogether. Of course this would be very different if you were using e.g. Java.

use-cases/caloinn/src/XMLHandler.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def __init__(self, particle_name, filename='binning.xml'):
+
+        tree = ET.parse(filename)
+        root = tree.getroot()
+
+        self.r_bins = []
+        self.a_bins = []
+        self.nBinAlphaPerlayer = []
+        self.alphaListPerLayer = []
```
jarlsondre, line 20: Python convention is to use snake_case, not camelCase

use-cases/caloinn/src/XMLHandler.py
```diff
@@ -0,0 +1,149 @@
...
+
+        self.eta_all_layers = []
+        self.phi_all_layers = []
+
+        self.SetEtaAndPhiFromPolar()
+        self.bin_edges = [0]
+        for i in range(len(self.bin_number)):
+            self.bin_edges.append(self.bin_number[i] + self.bin_edges[i])
+
+    def ReadPolarCoordinates(self, subelem):
```
jarlsondre, line 55: snake_case

use-cases/caloinn/src/binned.py
```diff
@@ -0,0 +1,224 @@
...
+            dims_c = []
+        if parameter_counts is None:
+            parameter_counts = {}
+
+        super().__init__(dims_in, dims_c, clamp=0.0, clamp_activation=lambda u: u, split_len=split_len)
+
+        assert bins >= 1, "need at least one bin"
+        assert all(s >= 0 for s in min_bin_sizes), "minimum bin size cannot be negative"
+        assert default_domain[1] > default_domain[0], "x domain must be increasing"
+        assert default_domain[3] > default_domain[2], "y domain must be increasing"
```
jarlsondre, line 43: Generally it is preferred to use `raise YourErrorHere` instead of `assert` outside of tests due to the semantics of an `AssertionError` vs a more specialized error. If you're unsure, you can usually use `ValueError`.

Issue #162, opened 2024-06-19
First Radio Astronomy use case integration


matbun:
Data is on the data lake.
Code is on a VM connected to the data lake.

PR #358, opened 2025-03-31
Addition of radio-astronomy use-case

okrochak:
# Summary

Includes documentation, test suite composed of 5 tests, and interface implementation to 
ensure smooth integration of the radio-astronomy source code into itwinai framework.

The training pipeline is quite sophisticated, as the model is composed of 3 subnetworks.
However, execution is simplified via config.yaml file. 

For more information, please view the related documentation for this use-case

Commit history can be quite messy as I was not familiar with itwinai at the beginning.
---

**Related issue : [https://github.com/interTwin-eu/itwinai/issues/162](url) **


docs/use-cases/radio-astronomy.rst
```diff
@@ -0,0 +1,52 @@
...
+Please visit the original repository for more technical information on the code. 
+This use-case features a sophisticated pipeline composed of few neural networks.
+
+Integration Author: Oleksandr Krochak, FZJ
+
+Environment Management
+-----------------------------------------------------------------------------------------------
+It is recommended to use UV environment for running this pipeline. 
+The overview of itwinai-wide module dependencies can be found in `intertwin/pyproject.toml`.
+By running `uv sync --extra devel --extra torch --extra radio-astronomy`, uv lockfile will
```
jarlsondre, line 14: "the `uv` lockfile"

docs/use-cases/radio-astronomy.rst
```diff
@@ -0,0 +1,52 @@
...
+This use-case features a sophisticated pipeline composed of few neural networks.
+
+Integration Author: Oleksandr Krochak, FZJ
+
+Environment Management
+-----------------------------------------------------------------------------------------------
+It is recommended to use UV environment for running this pipeline. 
+The overview of itwinai-wide module dependencies can be found in `intertwin/pyproject.toml`.
+By running `uv sync --extra devel --extra torch --extra radio-astronomy`, uv lockfile will 
+be generated/updated that ensures that correct dependencies are installed. If want to
```
jarlsondre, line 15: "If you want to"

docs/use-cases/radio-astronomy.rst
```diff
@@ -0,0 +1,52 @@
...
+It is recommended to use UV environment for running this pipeline. 
+The overview of itwinai-wide module dependencies can be found in `intertwin/pyproject.toml`.
+By running `uv sync --extra devel --extra torch --extra radio-astronomy`, uv lockfile will 
+be generated/updated that ensures that correct dependencies are installed. If want to 
+change some use-case specific dependencies, please do so in pyproject.toml in the radio-astronomy
+section. Afterwards, re-run `uv sync` with the same flags.
+
+Running from a configuration file
+-----------------------------------------------------------------------------------------------
+You can run the full pipeline sequency by executing the following commands locally.
```
jarlsondre, line 21: "sequency" -> "sequence" or "sequentially" I imagine?

use-cases/radio-astronomy/batch.sh
```diff
@@ -0,0 +1,57 @@
...
+
+srun_torchrun() {
+  srun --cpu-bind=none --ntasks-per-node=1 bash -c "torchrun \
+    --log_dir='logs' \
+    --nnodes=${SLURM_NNODES} \
+    --nproc_per_node=${SLURM_GPUS_PER_NODE} \
+    --rdzv_conf=is_host=\$(((SLURM_NODEID)) && echo 0 || echo 1) \
+    --rdzv_backend=c10d \
+    --rdzv_id=${SLURM_JOB_ID} \
+    --rdzv_endpoint=\$(scontrol show hostnames \"\$SLURM_JOB_NODELIST\" | head -n1):54123 \
```
jarlsondre, line 47: It might be cleaner to use `MASTER_PORT` and `MASTER_ADDR` here as variables, though not super important.

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,345 @@
...
+                                               (set False when running 'main.py' directly after)
+        """
+        super().__init__(name)
+        self.save_parameters(**self.locals2params(locals()), pop_self=False)
+
+        # TODO find a smart way to compute the right value for num_cpus
+
+        if not (os.path.exists(param_root) and os.path.exists(payload_root)):
+            os.makedirs(param_root, exist_ok=True)
+            os.makedirs(payload_root, exist_ok=True)
```
jarlsondre, line 52: Is there a good reason to handle both paths at the same time here, instead of separately? Also, since you're checking both and potentially calling `makedirs()` on something that doesn't exist, why not just remove the if statement altogether? You have `exist_ok=True` anyway, so you don't really need the if here.

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,345 @@
...
+        mask_tag: str,
+        image_directory: str,
+        mask_directory: str,
+        # Re-use some arguments to initialize PrepareFreqTimeImage class inside this class
+        do_rot_phase_avg: bool = True,
+        do_resize: bool = True,
+        resize_size: Tuple = (128,128),
+        mask_binarize_func: str = "thresh",
+        mask_maker_engine: Optional[dict] = None,
+        image_engine: PrepareFreqTimeImage = PrepareFreqTimeImage(
```
jarlsondre, line 81: In general, setting the default value to an instantiated mutable object is considered an anti-pattern in python, because `PrepareFreqTimeImage` is created once, when the function is defined. This means that any instances of this dataset will actually share the same instance of `PrepareFreqTimeImage`. This can lead to some real nasty bugs that are very very hard to find, so should be avoided unless absolutely necessary. Example: 

```python
>>> class Jarl:
...   def __init__(self, my_dict = {}):
...     self.my_dict = my_dict
...
>>> j = Jarl()
>>> j.my_dict["hei"] = True
>>> k = Jarl()
>>> k.my_dict
{'hei': True}
>>>
```

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,345 @@
...
+        self._mask_engine = PrepareFreqTimeImage(
+                    do_rot_phase_avg=True,
+                    do_binarize=True,
+                    do_resize=True,
+                    resize_size=resize_size,
+                    binarize_engine = BinarizeToMask(binarize_func=mask_binarize_func)
+                    )
+        
+        self._image_engine = image_engine
+        self._mask_engine = mask_engine
```
jarlsondre, line 114: What is going on with the parameter names here? You have `_img_engine` and `_image_engine` and these are different, but then you also make `_mask_engine` twice, where the second time you override the first one? Is this a mistake?

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,345 @@
...
+            if mask_maker_engine["model"] == "UNet":
+                mme_model = UNet()
+            else:
+                raise ValueError("Uknown model type for mask_maker_engine")
+            self._mask_maker_engine = PipelineImageToMask(
+                image_to_mask_network=mme_model,
+                trained_image_to_mask_network_path=mask_maker_engine["model_path"],
+            )
+        else:
+            self._mask_maker_engine = None
```
jarlsondre, line 127: Since this is at the end of the function, you could rather do: 
```python
if mask_maker_engine is None:
  self._mask_maker_engine = None
  return

# The rest of the stuff in the if, but now without the indentation
...
```
This allows the code to be a bit more readable

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,345 @@
...
+
+    def loadImagePair(self, index) -> ImageMaskPair:
```
jarlsondre, line 129: Not super important, but usually you'd wanna avoid using names of built-in functions like "index", "sum" etc. Of course you don't seem to be using it in this scope, but it's often nice to just avoid these kinds of things altogether.

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,345 @@
...
+        return img.float(), mask.float()
+    
+    def __get_descriptions__(self, index):
+        return self.loadImagePair(index).descriptions
+
+    def __len__(self):
+        search_pattern = os.path.join(self._image_directory, self._image_tag)
+        matching_files = glob.glob(search_pattern)
+        num_items = len(matching_files)
+        return num_items
```
jarlsondre, line 166: This function seems to calculate the length every time using `glob` etc., but the value should stay the same every time right? Wouldn't it make more sense to calculate this just once, in which case? I suppose you could also use `@cachedproperty`, as I believe this does this automatically.

use-cases/radio-astronomy/data.py
jarlsondre: Also, you generally want to avoid using camelCase in Python, so this should probably be `load_image_pair`, right?

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,345 @@
...
+        Finds all pickled files in the root folder, then splits them into
+        training, validation, and test sets based on the specified proportions.
+
+        Returns:
+            Tuple[Dataset, Dataset, Dataset]: Training, validation, and test datasets.
+        """
+
+        # Split file paths into train, validation, and test sets
+        generator = torch.Generator().manual_seed(self.rnd_seed)
+        [train_dataset, validation_dataset, test_dataset] = random_split(
```
jarlsondre, line 258: You don't really need the square brackets here, I think. You should be able to do `train_dataset, validation_dataset, test_dataset = ...`

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,345 @@
...
+
+class pipelinePulsarInterface(PipelineImageToFilterDelGraphtoIsPulsar):
+    def execute(self) -> PipelineImageToFilterDelGraphtoIsPulsar:
+        return self
+    
+class pipelineLabelsInterface(PipelineImageToFilterToCCtoLabels):
+    def execute(self) -> PipelineImageToFilterToCCtoLabels:
+        return self
+
+class testSuite:
```
jarlsondre, line 273: class names should generally start with a capital letter

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,345 @@
...
+            self.DelGraphtoIsPulsar = PipelineImageToFilterDelGraphtoIsPulsar(
+                image_to_mask_network,
+                trained_image_to_mask_network_path,
+                mask_filter_network,
+                trained_mask_filter_network_path,
+                signal_to_label_network,
+                trained_signal_to_label_network   
+            )
+
+            self.ToCCtoLabels = PipelineImageToFilterToCCtoLabels(
```
jarlsondre, line 301: snake_case

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,345 @@
...
+            image_data_set=data_subset,
+            image_label_set=data_label_subset,
+            plot_randomly=True,
+            batch_size=2
+        )
+
+        plt.show()
+        for i in plt.get_fignums():
+            fig = plt.figure(i)
+            fig.savefig(f"plots/figure_{i}.png")
```
jarlsondre, line 333: This could lead to an error if the `plots` directory doesn't exist, so you might want to do `mkdir` at some point here. It would be best practice to use `Path` from `pathlib` to do this

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,345 @@
...
+            
+        return print("Test Suite executed")       
+
+class ModelSaver:
+    def execute(self, model, path) -> None:
+        m_dict = model.state_dict()
+        # ensure correct saving syntax expected by the loader
+        m_new = OrderedDict([(k.replace("module.",""), v) if k.startswith("module") else (k, v) for k, v in m_dict.items()])
+        del m_dict
```
jarlsondre, line 342: Is there a specific reason for deleting this explicitly? Once the program goes out of the scope of the function it should automatically be deleted, right?

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,345 @@
...
+        torch.save(m_new, path)
+        print(f"Model saved at {path}")
+        return None
```
jarlsondre, line 345: This `return None` can be omitted

use-cases/radio-astronomy/main.py
```diff
@@ -0,0 +1,187 @@
...
+data_subset = data[offset+1:offset+size_of_set,:,:]
+data_label_subset = data_label[offset+1:offset+size_of_set]
+
+while(True):
+    ppl1f.test_on_real_data_from_npy_files(image_data_set=data_subset,image_label_set=data_label_subset,plot_details=True,plot_randomly=True,batch_size=2)
+    continue_loop = not(input('[y/n] to quit: ')=='y')    
+    plt.show()
+    if continue_loop==False:
+        break
+    plt.close()
```
jarlsondre, line 181: You don't really need the parentheses around `True` here. Also right now you always do `plt.show()` one more time after the user says "no", which might not be intended. 

Also, it is usually nice to avoid `while True` statements in Python if it is possible to know when it should end. I suppose it might be hard in your case since you're doing two things in the loop, though, but would generally be cleaner to structure your loop to use the `continue_loop` variable instead of `break`, if possible.

use-cases/radio-astronomy/src/pulsar_analysis/information_packet_formats.py
```diff
@@ -0,0 +1,95 @@
...
+    
+    def add_flux(self,radio_packet:list[list[float]]):
+        try:
+            freqs_received = radio_packet[1].tolist()            
+        except:
+            freqs_received = radio_packet[1]            
+        try:
+            flux_received = radio_packet[0].tolist()
+        except AttributeError:
+            flux_received = radio_packet[0]
```
jarlsondre, line 51: The type hint seems to suggest that the elements of `radio_packet` should be lists. If this is the case, why is your `try/catch` necessary?

use-cases/radio-astronomy/src/pulsar_analysis/information_packet_formats.py
```diff
@@ -0,0 +1,95 @@
...
+    
+    def write_payload_to_jsonfile(self,file_name:str):
+        with open(file_name, "w") as outfile:
+            json_object = json.dumps(self.__dict__,indent=4) 
+            outfile.write(json_object)
+
+    @classmethod
+    def read_payload_from_jsonfile(cls,filename:str):
+        cls_obj = cls(freqs=[])
+        f = open(filename)
```
jarlsondre, line 82: I see that you're not using a context manager here, i.e. `with open(...) as f`, but you also never seem to close the file. You should probably close the file before returning to avoid any problems with the I/O.

use-cases/radio-astronomy/src/pulsar_analysis/neural_network_models.py
```diff
@@ -0,0 +1,554 @@
...
+
+class CustomLossUNet(nn.Module):
+    def __init__(self):
+        super(CustomLossUNet, self).__init__()
+
+    def forward(self, output, target):        
+        loss = (output - target) ** 2
+        mask = target >= 0
+        high_cost = (loss * mask.float()).mean()
+        return high_cost
```
jarlsondre, line 326: Probably not a problem the way it is right now, but are there no built-in loss functions that can handle these computations? That way you could be certain that it is optimized fully. Again, probably okay the way it is though.

use-cases/radio-astronomy/src/pulsar_analysis/neural_network_models.py
```diff
@@ -0,0 +1,554 @@
...
+
+    def forward(self, preds, targets):
+        #preds = (preds > 0.5).float()
+        preds = preds.contiguous()
+        targets = targets.contiguous()
+        
+        intersection = (preds * targets).sum(dim=2).sum(dim=2)
+        dice_coef = (2. * intersection + self.smooth) / (preds.sum(dim=2).sum(dim=2) + targets.sum(dim=2).sum(dim=2) + self.smooth)
+        
+        return 1 - dice_coef.mean()
```
jarlsondre, line 341: I'm fairly certain I have seen library implementations of the DICE score before, so could be better to use those to be certain that there are no bugs in your code.

use-cases/radio-astronomy/src/pulsar_analysis/pipeline_methods.py
```diff
@@ -0,0 +1,1577 @@
...
+
+    def __call__(
+        self, image: np.ndarray, return_bool: bool = True, return_steps: bool = False
+    ):
+        mask = self.image_to_mask_method(image=image)
+        signal = self.mask_to_signal_method(mask=mask)
+        label = self.signal_to_label_method(signal=signal)
+
+        if return_bool and not return_steps:
+            if label > 0.5:
```
jarlsondre, line 305: `return label > 0.5`

jarlsondre: The same goes for the below in this function as well

use-cases/radio-astronomy/src/pulsar_analysis/pipeline_methods.py
```diff
@@ -0,0 +1,1577 @@
...
+        )
+        self.__image_to_mask_network.eval()
+
+        self.__signal_to_label_network = self.__signal_to_label_network.to(self.device)
+        self.__signal_to_label_network.load_state_dict(
+            torch.load(trained_signal_to_label_network,map_location=torch.device(self.device),weights_only=True)
+        )
+        self.__signal_to_label_network.eval()
+
+    def __call__(
```
jarlsondre, line 297: This function can return a lot of different stuff and it makes it quite hard to understand what's going on, so you should consider splitting this into different functions or different objects, if possible.

use-cases/radio-astronomy/src/pulsar_analysis/pipeline_methods.py
```diff
@@ -0,0 +1,1577 @@
...
+                bool_val = True
+            else:
+                bool_val = False
+            return bool_val, mask, signal, label
+        elif not return_bool and return_steps:
+            return mask, signal, label
+        else:
+            return label
+
+    def image_to_mask_method(self, image: np.ndarray):
```
jarlsondre, line 320: The name `..._method` doesn't really add anything because how could it not be a method?

use-cases/radio-astronomy/src/pulsar_analysis/pipeline_methods.py
```diff
@@ -0,0 +1,1577 @@
...
+        """Method to convert mask to delaygraph and extract lags as signal
+
+        Args:
+            image (np.ndarray): mask
+
+        Returns:
+            np.ndarray: x_lags as signal
+        """
+        delay_graph_engine = DelayGraph()
+        x_lags, __ = delay_graph_engine(dispersed_freq_time=mask)
```
jarlsondre, line 352: Why the double underscore here?

use-cases/radio-astronomy/src/pulsar_analysis/pipeline_methods.py
```diff
@@ -0,0 +1,1577 @@
...
+        Args:
+            image_data_set (np.memmap): image dataset as numpy array
+            image_label_set (np.memmap | None, optional): label dataset as numpy array. Defaults to None.
+            plot_details (bool, optional): if True then plot the results. Defaults to False.
+            plot_randomly (bool, optional): If True then randomly choose images from dataset. Defaults to True.
+            batch_size (int, optional): number of images to test. minimum is 2. Defaults to 5.
+
+
+        """
+        if plot_details == False:
```
jarlsondre, line 483: `if not plot_details`

use-cases/radio-astronomy/src/pulsar_analysis/postprocessing.py
```diff
@@ -0,0 +1,690 @@
...
+            line_func,
+            y_coors,
+            x_coors,
+            maxfev=5000,
+            bounds=([-np.inf, -1], [np.inf, 1]),
+        )
+        m_fitted = popt[0]
+        return np.arctan(1 / m_fitted) * 180 / np.pi, line_func, popt
+
+    def __repr__(self):
```
jarlsondre, line 214: Any reason for the `pass` here?

use-cases/radio-astronomy/src/pulsar_analysis/postprocessing.py
```diff
@@ -0,0 +1,690 @@
...
+        """
+        # not_node_image = np.logical_not(node_pixels)
+        node_pixels.astype(int)
+        skeleton_img_no_nodes = skeleton_img
+        skeleton_img_no_nodes.astype(int)
+        skeleton_img_no_nodes[node_pixels == 1] = 0
+        return skeleton_img_no_nodes
+
+    @staticmethod
+    def label_skeleton(skeleton_img: np.ndarray):
```
jarlsondre, line 342: What's the point of this function? It just calls another function for you, so why not just call the other function directly?

use-cases/radio-astronomy/data.py
okrochak: Interesting, thanks for explaining this !

use-cases/radio-astronomy/data.py
okrochak: The dataset classes are quite messy, because i tried to re-factor some of the use-case code. However now I see it's still quite difficult. I'll write a new dataset class that can be used in 3 instances for 3 networks, with more explici initialization

okrochak:
I'm refactoring some of the use-case integration classes now based off Jarl's feedback. I will update this comment once the integration is done and the code is ready for review.

use-cases/radio-astronomy/data.py
okrochak: Yes, but I won't change it as it might break some other use-case code

use-cases/radio-astronomy/.config-test.yaml
```diff
@@ -0,0 +1,289 @@
...
+    training_step:
+      _target_: trainer.PulsarTrainer
+      strategy: ${strategy}
+      model:
+       _target_: src.pulsar_analysis.neural_network_models.UNet
+      loss:
+        _target_: src.pulsar_analysis.neural_network_models.WeightedBCELoss
+        pos_weight: 3
+        neg_weight: 1
+      epochs: 1
```
r-sarma, line 128: It is probably preferable to declare all the variables at the beginning of the config file - easier to edit.

use-cases/radio-astronomy/batch-jsc.sh
```diff
@@ -0,0 +1,62 @@
...
+}
+
+## Data generation - best to run on a single node without GPUs
+srun $(which itwinai) exec-pipeline +pipe_key=syndata_pipeline 
+
+## Newtork training and evaluation - can be executed on multi-node / multi-GPU
+srun_torchrun $(which itwinai) exec-pipeline +pipe_key=unet_pipeline
+srun_torchrun $(which itwinai) exec-pipeline +pipe_key=fcnn_pipeline
+srun_torchrun $(which itwinai) exec-pipeline +pipe_key=cnn1d_pipeline
+srun_torchrun $(which itwinai) exec-pipeline +pipe_key=evaluate_pipeline
```
r-sarma, line 62: So all of these pipelines are meant to be run sequentially?

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,374 @@
...
+        """Initialize the synthesizeData class.
+        Args:
+            name [optional] (str):  name of the data getter component.
+            param_root      (str):  folder where synthetic param data will be saved.
+            payload_root    (str):  folder where synthetic payload data will be saved.
+            tag             (str):  tag which is used as prefix for the generated files.
+            num_cpus        (int):  number of CPUs used for parallel processing.
+            num_payloads    (int):  number of generated examples.
+            plot            (bool): if True, plotting routine is activated \
+                                               (set False when running 'main.py' directly after)
```
r-sarma, line 44: No so important, but perhaps also add Returns None?

use-cases/radio-astronomy/main.py
```diff
@@ -0,0 +1,136 @@
...
+data_subset         = data[inputs.offset+1:inputs.offset+inputs.size_of_set,:,:]
+data_label_subset   = data_label[inputs.offset+1:inputs.offset+inputs.size_of_set]
+
+while(True):
+    ppl1f.test_on_real_data_from_npy_files(image_data_set=data_subset,image_label_set=data_label_subset,plot_details=True,plot_randomly=True,batch_size=2)
+    plt.show()
+    if (input('[y/n] to quit: ')=='y'):
+        break
+    plt.close()
+    ppl2f.test_on_real_data_from_npy_files(image_data_set=data_subset,image_label_set=data_label_subset,plot_randomly=True,batch_size=2)
```
r-sarma, line 132: Same as above

use-cases/radio-astronomy/main.py
r-sarma, line 1: Is this file used if a user doesn't want to use the pipeline?

use-cases/radio-astronomy/src/pulsar_analysis/pipeline_methods.py
```diff
@@ -0,0 +1,1577 @@
...
+
+        Returns:
+            np.ndarray: x_lags as signal
+        """
+        delay_graph_engine = DelayGraph()
+        x_lags, __ = delay_graph_engine(dispersed_freq_time=mask)
+        signal = x_lags.flatten()
+        return signal
+
+    def signal_to_label_method(self, signal: np.ndarray):
```
r-sarma, line 356: Agree with Jarl's comment above. In general, the method names are very long. I understand that these are probably from the use-case. But please considering shortening wherever feasible.

use-cases/radio-astronomy/trainer.py
```diff
@@ -0,0 +1,121 @@
...
+        self,
+        config:                 Dict | TrainingConfiguration | None             = None,
+        strategy:               Literal["ddp", "deepspeed", "horovod"] | None   = None,
+        logger:                 Logger | None                                   = None,
+        epochs:                 int                                             = 3,
+        model:                  nn.Module                                       = None,
+        loss:                   nn.Module                                       = None,
+        store_trained_model_at: str                                             = ".models/model.pt",
+        name:                   Optional[str]                                   = None,
+    ) -> None:
```
r-sarma, line 31: It is more of a styling issue, but I would not prefer large spaces here. It is more difficult to follow, imo.

use-cases/radio-astronomy/trainer.py
```diff
@@ -0,0 +1,121 @@
...
+# Created by: Oleksandr Krochak
+# --------------------------------------------------------------------------------------
+
+import os
+from typing import Any, Dict, Literal, Optional, Tuple
+from torch.utils.data import Dataset, DataLoader
+
+import torch.nn as nn
+from torch import device, cuda, save
+from itwinai.torch.trainer import TorchTrainer, RayTorchTrainer
```
r-sarma, line 12: Is RayTorchTrainer used?

use-cases/radio-astronomy/trainer.py
```diff
@@ -0,0 +1,121 @@
...
+from itwinai.torch.config import TrainingConfiguration
+from itwinai.loggers import EpochTimeTracker, Logger
```
r-sarma, line 14: Since you already import EpochTimeTracker, perhaps its a good idea to add some profiling and scaling stuff as well. That might be interesting for the use-case. Ofcourse not very important for now.

use-cases/radio-astronomy/src/pulsar_analysis/postprocessing.py
```diff
@@ -0,0 +1,690 @@
...
+            ax.set_ylabel("freq channel")
+            ax.grid("True")
+            ax.invert_yaxis()
+            ax.set_aspect("auto")
+
+    @classmethod
+    def return_detected_categories(
+        cls, labelled_skeleton: np.ndarray, return_detailed_results: bool = False
+    ):
+        """Method that summarizes the results into a dict of categories with values of each keys representing the mean length of the categories
```
r-sarma, line 583: Long sentence. Consider adding breaks?

use-cases/radio-astronomy/src/pulsar_analysis/postprocessing.py
```diff
@@ -0,0 +1,690 @@
...
+from scipy.optimize import curve_fit
+from scipy.signal import savgol_filter
+from scipy.ndimage import convolve, gaussian_filter1d
+from skimage.morphology import skeletonize, label, binary_dilation
+import matplotlib.pyplot as plt
+import matplotlib.patches as patches
+
+
+class DelayGraph:
+    """Class involving methods to generate delaygraph from freq-time image. DelayGraph is generated by calculating the lags with the mean signal with each of the freq channel signals"""
```
r-sarma, line 12: Consider adding line breaks.

use-cases/radio-astronomy/.config-test.yaml
```diff
@@ -0,0 +1,289 @@
...
+###  THIS IS A CONFIGURATION FILE FOR THE TEST SUITE   ###
+###     PLEASE DO NOT EDIT THIS FILE UNLESS WORKING    ###
+### ON THE TEST SUITE FOR THE RADIO-ASTRONOMY USE-CASE ### 
+
+
+# input variables
+
+num_payloads:     4
+learning_rate:    0.001
+batch_size:       4 #128
```
lineick, line 16: low batch size

use-cases/radio-astronomy/config.yaml
```diff
@@ -0,0 +1,282 @@
...
+      epochs: 10 #
+      store_trained_model_at: ${fcnn_st}
+      name: FilterCNNTrainer
+      config:
+        generator: simple
+        optimizer: adam
+        batch_size: ${batch_size}
+        optim_lr: ${learning_rate}
+        save_best: 1
+        shuffle_train: 1
```
lineick, line 184: are these bools or really ints?

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,374 @@
...
+    
+    def __get_descriptions__(self, index):
+        return self.load_image_pair(index).descriptions
+    
+    def __len__(self):
+        return self._len
+
+    def plot(self, index):
+
+        ##TODO: make this method HPC-friendy
```
lineick, line 222: Unresolved TODO?

use-cases/radio-astronomy/src/pulsar_analysis/information_packet_formats.py
```diff
@@ -0,0 +1,95 @@
...
+import json
+import matplotlib.pyplot as plt    
+
+class Payload:
+    def __init__(self,freqs:list[float],bandwidths:list[float]=None):
+        freqs = [f for f in freqs]        
+        self.freqs = freqs
+        self.dataframe = []
+        self.bandwidths = bandwidths
+        self.description = {'Pulsar': None, 'NBRFI': None, 'BBRFI': None}
```
lineick, line 12: missing docstring

use-cases/radio-astronomy/src/pulsar_analysis/information_packet_formats.py
```diff
@@ -0,0 +1,95 @@
...
+            plt.gca().set_aspect('auto')
+            return plt.gca()
+
+            
+        pass
+    
+    def add_flux(self,radio_packet:list[list[float]]):
+        try:
+            freqs_received = radio_packet[1].tolist()            
+        except:
```
lineick, line 46: except AttributeError?

use-cases/radio-astronomy/src/pulsar_analysis/neural_network_models.py
```diff
@@ -0,0 +1,554 @@
...
+from collections import OrderedDict
+
+
+class UNet(nn.Module):
+    """_summary_
+
+    :param nn: _description_
+    :type nn: _type_
+    """
+
```
lineick, line 14: empty docstring

use-cases/radio-astronomy/src/pulsar_analysis/neural_network_models.py
```diff
@@ -0,0 +1,554 @@
...
+                ]
+            )
+        )
+    
+class UNetFilter(nn.Module):
+    """_summary_
+
+    :param nn: _description_
+    :type nn: _type_
+    """
```
lineick, line 204: empty docstring

use-cases/radio-astronomy/src/pulsar_analysis/pipeline_methods.py
```diff
@@ -0,0 +1,1577 @@
...
+        """
+        image = (
+            torch.tensor(image, requires_grad=False).unsqueeze(0).unsqueeze(0).float()
+        )
+
+
+        ## Original code
+        # with torch.no_grad():
+        #     pred = self.__image_to_mask_network(image.to(self.device))
+        # pred = pred.to("cpu")
```
lineick, line 234: remove

use-cases/radio-astronomy/src/pulsar_analysis/pipeline_methods.py
```diff
@@ -0,0 +1,1577 @@
...
+            ax[i, 1].imshow(mask_current)
+            ax[i, 1].set_title(f"idx: {idx}:{given_image_label_current}",fontsize=10)
+
+            ax[i, 2].imshow(pred_binarized)
+            ax[i, 2].set_title(f"Segmented Mask")
+
+            if results:
+                # fig1, ax = plt.subplots(1, 1, figsize=(5, 5))
+                ax[i, 3].imshow(image_current)
+                for result in results:
```
lineick, line 1028: invert if statement

if not results: pass

use-cases/radio-astronomy/src/pulsar_analysis/pipeline_methods.py
```diff
@@ -0,0 +1,1577 @@
...
+            results, pred_binarized, labelled_skeleton = self.__call__(
+                image=image_current, return_steps=True
+            )
+            ax[i, 0].imshow(image_current)
+            ax[i, 0].set_title(f"idx: {idx}:{given_image_label_current}")
+
+            ax[i, 1].imshow(pred_binarized)
+            ax[i, 1].set_title(f"Segmented Mask")
+
+            if results:
```
lineick, line 1132: invert if

use-cases/radio-astronomy/src/pulsar_analysis/pipeline_methods.py
```diff
@@ -0,0 +1,1577 @@
...
+            ax[i, 1].imshow(mask_current)
+            ax[i, 1].set_title(f"Original Mask",fontsize=10)
+
+            ax[i, 2].imshow(pred_binarized)
+            ax[i, 2].set_title(f"Segmented Mask")
+
+            ax[i, 3].imshow(pred_binarized_filtered)
+            ax[i, 3].set_title(f"Filtered Mask")
+
+            if results:
```
lineick, line 1407: if inversion

use-cases/radio-astronomy/src/pulsar_analysis/pipeline_methods.py
```diff
@@ -0,0 +1,1577 @@
...
+            ax[i, 0].imshow(image_current)
+            ax[i, 0].set_title(f"idx: {idx}:{given_image_label_current}")
+
+            ax[i, 1].imshow(pred_binarized)
+            ax[i, 1].set_title(f"Segmented Mask")
+
+            ax[i, 2].imshow(pred_binarized_filtered)
+            ax[i, 2].set_title(f"Filtered Mask")
+
+            if results:
```
lineick, line 1516: if inversion

use-cases/radio-astronomy/main.py
okrochak: So you can run the pipeline either through this file or via CLI with `itwinai exec-pipeline ...`. If CLI option is used, this and inputs.py files are not necessary

tests/loggers/test_prov4ml.py
```diff
@@ -16,13 +16,13 @@ def logger_instance():
 @pytest.fixture
 def mlflow_run():
     with tempfile.TemporaryDirectory() as mlflow_temp_dir:
+        mlflow.end_run() # Ensure no active run
         mlflow.set_tracking_uri(Path(mlflow_temp_dir).resolve().as_uri())
         experiment_id = mlflow.create_experiment("temporary_experiment")
         mlflow.set_experiment(experiment_id=experiment_id)
         yield mlflow.start_run()
         mlflow.end_run()
```
matbun, line 24: I guess you can remove this

matbun (changes requested):

use-cases/radio-astronomy/trainer.py
```diff
@@ -0,0 +1,124 @@
...
+        distribute_kwargs = self.get_default_distributed_kwargs()
+
+        # Distributed model, optimizer, and scheduler
+        (self.model, self.optimizer, self.lr_scheduler) = self.strategy.distributed(
+            self.model, self.optimizer, self.lr_scheduler, **distribute_kwargs
+        )
+
+    def write_model(self) -> None:
+        """Write the model to disk."""
+        save(self.model.state_dict(), self.store_trained_model_at)
```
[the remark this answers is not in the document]
okrochak: it was present in the original use-case trainer so I'd like to keep it

use-cases/radio-astronomy/.test_dataset/train_v0_0_payload_detected.json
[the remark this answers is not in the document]
okrochak: I need this for testing purposes. How should I go on about it?

use-cases/radio-astronomy/trainer.py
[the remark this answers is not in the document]
matbun: ok

tests/use-cases/test_radio-astronomy.py
jarlsondre, line 1: Is there a particular reason for the underscore followed by a hyphen in the file name or was this an accident?

use-cases/radio-astronomy/.pytest-clean
jarlsondre, line 1: Is there a good reason for writing this in bash as opposed to using python commands to do the same? I imagine it might be easier to check for the existence of the different files etc if you did it using `pathlib` or `os` or something, and you could just store this in a function, making it easier for users to look up.

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,414 @@
...
+            model_name = engine_settings["mask_maker"].pop("model")
+            if model_name == "UNet":
+                mme_model = UNet()
+            else:
+                raise ValueError(
+                    "Uknown model type in engine_settings['mask_maker']['model']"
+                )
+            self._mask_maker_engine = PipelineImageToMask(
+                **engine_settings["mask_maker"], image_to_mask_network=mme_model
+            )
```
jarlsondre, line 177: This could be a little more neat if you did the following:

```python
if self._type != "filtercnn":
  return
model_name = engine_settings ...
if model_name != "UNet":
  raise ...

self._mask_maker_engine = ...
```
This way you avoid any indentation and you also handle data validation "separately", i.e. instead of checking if it is correct and then doing `raise` under the `else`, you would check if it is wrong directly. Just makes more sense (to me at least) semantically when reading the code.

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,414 @@
...
+                label_vector = [1]
+            else:
+                label_vector = [0]
+
+            label_vector = torch.tensor(label_vector, dtype=torch.float32)
+            pair = torch.tensor(signal, dtype=torch.float32).unsqueeze(0), label_vector
+
+        return pair
+
+    def __getitem__(self, item_id: int):
```
jarlsondre, line 223: What is the type of the returned element here?

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,414 @@
...
+
+    def __get_descriptions__(self, item_id: int):
+        """Provide descriptions of the data point with the given item_id."""
+        return self.load_image_pair(item_id).descriptions
+
+    def __len__(self):
+        """Return the length of the dataset pre-computed during the initialization."""
+        return self._len
+
+    def plot(self, item_id: int):
```
jarlsondre, line 246: This method only seems to have a `return` statement under one of the conditional branches. Is this intended?

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,414 @@
...
+            batch_size=2,
+        )
+
+        # plt.show()
+        os.makedirs("plots", exist_ok=True)
+        for i in plt.get_fignums():
+            fig = plt.figure(i)
+            fig.savefig(f"plots/figure_{i}.png")
+
+        return print("Test Suite executed")
```
jarlsondre, line 400: Wouldn't this essentially return `None`?

tests/use-cases/test_radio-astronomy.py
okrochak: The reason was that all the tests begin with test_, but the branch name is radio-astronomy. But I'll change it to double underscore, I agree that looks better.

use-cases/radio-astronomy/inputs.py
jarlsondre, line 1: Why not use a `.yaml` file instead?

use-cases/radio-astronomy/.pytest-clean
okrochak: I think bash is easier for most people working with HPC

use-cases/radio-astronomy/data.py
okrochak: The model would still need to be initialized, so you have indentation either way. I think this is fine as is

use-cases/radio-astronomy/data.py
```diff
@@ -0,0 +1,414 @@
...
+    def __getitem__(self, item_id: int):
+        """Use load_image_pair method to retrieve the data point with the given item_id.
+        Output is network architecture dependent.
+        Returns either image-mask pair (UNET, FCNN) or signal-label pair (CNN1D)."""
+
+        ### Implementation is slightly different per architecture
+        ### due to the original use-case code
+        if self._type == "cnn1d":
+            return self.load_image_pair(item_id)
+        else:
```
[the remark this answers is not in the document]
okrochak: I guess not, but I believe this makes it clearer that there are two various methods to return. I've added type check in the 2nd clause as well to be more explicit

use-cases/radio-astronomy/inputs.py
okrochak: Basically this is the same as a config.yaml file. The only reason I add this `inputs.py` is that in the original use-case code, there was a main script with some pre-defined variables. So this is to keep the original functionality

.github/workflows/pytest.yml
okrochak, line 1: I suggest adding a '--dist loadfile' option to the pytest call. This will ensure that tests from 1 file are executed by the same worker, avoiding potential bugs that are not consistently reproducible.

.github/workflows/pytest.yml
```diff
@@ -42,35 +42,7 @@ jobs:
...
-  #     #   shell: bash -l {0}
-  #     #   run: make tensorflow-env-cpu
-
-  #     # NOTE, to change the name of the env in which tests are run, set custom TORCH_ENV
-  #     # and TF_ENV env variables. Default environment names are ".venv-pytorch" and
-  #     # ".venv-tf"
-
-  #     - name: Run pytest for workflows
-  #       shell: bash -l {0}
-  #       run: .venv-pytorch/bin/pytest -v -n logical ./tests/ -m "not hpc and not tensorflow"
```
matbun, line 76: I would leave this commented out just because the Dagger pipeline we currently use is still "experimental", as we haven't been using it for long time and I guess I'm the only one familiar with Dagger. The commented out code at least gives a fallback in case something goes wrong with Dagger

tests/use-cases/test_radio_astronomy.py
matbun, line 1: To be consistent with the docstring style in itwinai, please avoid newline after `"""`.

Example: do this

```
"""Returns absolute path to torch virtual environment."""
```

Rather than this:

```
"""
Returns absolute path to torch virtual environment.
"""
```

use-cases/radio-astronomy/.config-test.yaml
matbun, line 1: Is this file really needed? Wouldn't it be easier to reuse `config.yaml` in the tests? If you need some specific/custom pipelines for the tests, you could add them in `config.yaml`

okrochak: No, because the tests are set up with arguments that lead to minimum resource use during the execution.

use-cases/radio-astronomy/data.py
okrochak: Added the type

use-cases/radio-astronomy/data.py
okrochak: I believe this was use-case's original idea, but I've removed the return of axes. Now it returns None in both cases.

tests/use-cases/test_radio_astronomy.py
okrochak: Changed everywhere to be consistent

use-cases/radio-astronomy/data.py
okrochak: Yeah, removed it

matbun (approved):

use-cases/radio-astronomy/batch-jsc.sh
matbun, line 1: would it be possible to have examples on how to launch distributed training also with horovod and deepspeed?

matbun (changes requested):

use-cases/radio-astronomy/batch-jsc.sh
okrochak: Isn't it just switching the strategy in config.yaml? I've tried it and it seemed to work for me

okrochak: I've tried to build horovod, but I'm encountering some difficulties. There is an example of how to run horovod command in lattice-qcd use-case, so that's an option if a user wants to run with that, but to be honest I don't think providing support to running with horovod should be a priority as deepspeed and torch-ddp are much more actively maintained.

PR #362, opened 2025-04-09
Integration of Lattice QCD use-case

r-sarma:
This PR integrates the Lattice-QCD use-case.
Integration covers:
- Distributed strategy based on itwinai `TorchTrainer`
- Configuration file based execution of training
- Integration of logging and profiling functionalities


docs/use-cases/latticeqcd_doc.rst
```diff
@@ -0,0 +1,34 @@
+Normalizing flow for generating lattice field configurations (Lattice QCD, ETHZ/CSIC)
+=====================================================================================
+
+The code is adapted from `this notebook <https://github.com/jkomijani/normflow_>`_ from the Lattice QCD use-case.
```
jarlsondre, line 4: "use case" should only be hyphenated if it acts as an adjective, but in this case it is the subject and should thus remain without a hyphen.

docs/use-cases/latticeqcd_doc.rst
```diff
@@ -0,0 +1,34 @@
...
+
+More information on the use-case is available in the published deliverables,
```
jarlsondre, line 6: Same with the hyphen in "use case" as above.

docs/use-cases/latticeqcd_doc.rst
```diff
@@ -0,0 +1,34 @@
...
+With ``uv``, the ``normflow`` repository (located under ``use-cases/lattice-qcd``) can be installed
+using the ``--extra lattice-qcd`` flag. For example, to install the normflow environment in itwinai with
+Torch support in development mode, run:
+
+.. code-block:: bash
+
+   uv sync --no-cache --extra dev --extra torch --extra lattice-qcd
+
+This will install all required dependencies for both ``itwinai`` and ``normflow``, including
+Torch and optional developer tools.
```
jarlsondre, line 27: Super minor thing: I would personally either say "PyTorch" instead of "Torch" or say `torch` with backticks here.

pyproject.toml
```diff
@@ -84,6 +84,13 @@ hpo = [
   "hpbandster>=0.7.0",
   "gpy>=1.13.2",
 ]
+# Dependencies for Lattice-QCD
+lattice-qcd = ["normflow"]
+
+[tool.hatch.metadata]
+allow-direct-references = true
+
+
```
matbun, line 93: as for the gitignore, I'd suggest moving use case-specific information (in this case the dependencies) in the use case folder, without touching the upstream pyproject.toml. A use case-specific pyproject.toml can be created under the use case folder, or in the plugin repo

matbun (changes requested):
I notice that many files appear to be copied from the normflow repository. Beyond the implications related to licensing and proper acknowledgment of the original developers, this approach makes it difficult for us to review and clearly understand the actual changes resulting from the integration. I can see the logic of this method for a plugin repository (which we already have), but for itwinai, wouldn’t it be simpler to install normflow (perhaps from a specific commit ID) and push only the files that have been modified during the integration?

use-cases/lattice-qcd/README.md
```diff
@@ -0,0 +1,262 @@
...
+be accessed [here](https://github.com/r-sarma/normflow_itwinai_source). The files
+which are part of the itwinai integration for `normflow` are included in the `itwinai`
+repository. In order to correctly build the environment for `normflow`, it is
+important to follow these steps mentioned below.
+
+1. Once `itwinai` is cloned and the environment is setup (see instructions
+[here](https://itwinai.readthedocs.io/latest/installation/developer_installation.html)),
+please navigate to the `use-cases/lattice-qcd` directory:
+
+```
```
jarlsondre, line 54: For these code blocks, you'd be better off using ` ```bash`. That way you can get syntax highlighting. This goes for all the blocks, really

use-cases/lattice-qcd/README.md
```diff
@@ -0,0 +1,262 @@
...
+cd use-cases/lattice-qcd
+```
+
+2. In this directory, the source files from the integration are included. The
+other files required to build the `normflow` environment are imported from
+the forked repository mentioned above. To do this, please run the following lines
+on your terminal:
+
+```
```
jarlsondre, line 63: Syntax highlighting (like above)

use-cases/lattice-qcd/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+These lines clone the required files from the forked repository, and then
+removes the `tmp-repo` since this is not needed for setting up the environment.
+
+3. Next, the file `_normflowcore.py` has to be moved to the newly-created
+`src` directory. This file contains the `Trainer` configuration and enables
+access to all the logic defined for distributed training strategy, logging,
+profiling and other functionalities offered by `itwinai`.
+
+```
```
jarlsondre, line 84: Syntax highlighting (like above)

use-cases/lattice-qcd/README.md
```diff
@@ -0,0 +1,262 @@
...
+    action:
+        _target_: normflow.action.ScalarPhi4Action
+        kappa: 0
+        m_sq: -1.2
+        lambd: 0.5
+```
+
+In this example, we have:
+
+-   **Prior Distribution**: A normal distribution is used with a shape of`[1]`.
```
jarlsondre, line 121: I am not quite sure what "shape of `[1]`" means here. Do you mean in the sense that you use a univariate normal distribution? I.e. that it is a scalar? In which case, I am more used to the notation `(1,)`, but I think I prefer "scalar" or "univariate".

use-cases/lattice-qcd/README.md
```diff
@@ -0,0 +1,262 @@
...
+    `kappa=0`, `m_sq=-2.0`, and `lambda=0.2`.
+
+-   **Neural Network**: The `DistConvertor_` class is used to create the
+    transformation network, with `knots_len=10` and symmetry enabled.
+    Any instance of this class converts the probability distribution of inputs
+    using a rational quadratic spline. In this example, the spline has 10 knots,
+    and the distribution is assumed to be symmetric with respect to the origin.
+
+The other parameters for training are specified within the same `yaml` file.
+For conciseness, only few of the options, the `config` parameter, epochs and strategy
```
jarlsondre, line 133: This should be "only a few" or "only some of the options", right?

use-cases/lattice-qcd/README.md
```diff
@@ -0,0 +1,262 @@
...
+#################################
+```
+
+This output indicates the loss values at specified epochs during the training
+process, providing insight into the model's performance over time.
+
+Alternatively, the model initialization and training can also be performed
+using the `train.py` file, which provides another interface to the user to build
+the models and launch the training with configurations files defined within the
+Python script. When using this script to launch the training, one can do so by:
```
jarlsondre, line 197: [nitpick] Maybe instead say "One can use this script to launch the training like this:" or something? The current formulation just sounded a bit weird, but I think it is understandable so you don't really have to change this

use-cases/lattice-qcd/README.md
```diff
@@ -0,0 +1,262 @@
...
+```
+python train.py
+```
+
+For working on HPC systems,  a `startscript.sh` file is provided. This can be launched
```
jarlsondre, line 202: I could be wrong here, but I haven't seen the term HPC introduced in the text yet, so could be nice to write it out? Or maybe it is assumed as common knowledge?

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+        Args:
+            batch_size (int, optional): Number of samples to draw. Defaults to 1.
+            preprocess_func (Callable, optional): A function to adjust the prior
+                samples if needed. It should take samples and log probabilities as
+                input and return modified values.
+
+        Returns:
+            Tuple[torch.Tensor, torch.Tensor]:
+                - `y`: Generated samples.
+                - `logq`: Log probabilities of the samples.
```
jarlsondre, line 127: Personally I prefer doing the following for brevity when returning built-in tuples:
```python
"""
Returns:
  torch.Tensor: Generated Samples
  torch.Tensor: Log probabilities of the samples
"""
```
Definitely not required to do, but I think it makes it a bit easier to read.

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+
+        if preprocess_func is not None:
+            x, logr = preprocess_func(x, logr)
+
+        y, logj = self._model.net_(x)
+        logq = logr - logj
+        return y, logq
+
+    @torch.no_grad()
+    def sample__(
```
jarlsondre, line 139: I suppose this is part of Javad's convention, but I am not a fan of using double underscores like this. I would rather just change the name to something like `sample_with_log` or `sample_log` or something. Makes it more readable IMO.

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+        """Sets up random seed for each worker in a distributed setting.
+        This methods ensures that each workers receives a unique seed. The main worker
+        generates the seeds and broadcasts to other workers.
+
+        Args:
+            rank (int): The rank of the current worker.
+            world_size (int): The total number of workers in the distributed setting
+
+        Returns:
+            None
```
jarlsondre, line 223: I think you're fine to omit the "Returns:" section if it doesn't return anything.

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+
+        # Broadcast seed list to all workers
+        seeds_torch_list = self.strategy.allgather_obj(seeds_torch)
+        seeds_torch = next(s for s in seeds_torch_list if s is not None)
+
+        # Set seed for current worker based on its rank
+        seed = seeds_torch[rank]
+        torch.manual_seed(seed)
+
+        sys.stdout.write(f"Rank {rank} has been assigned seed {seed}\n")
```
jarlsondre, line 240: What is the reason for using `sys.stdout` here instead of `print()` or a logger? In one of the recent updates to `itwinai` we actually added a `py_logger` that is meant to serve as a replacement to prints but probably also this. You can have a look at `cli.py` in the `main` branch to see an example of how we do this.

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+
+        Args:
+            train_dataset (Dataset): training dataset object.
+            validation_dataset (Dataset | None): validation dataset object.
+                Defaults to None.
+            test_dataset (Dataset | None): test dataset object.
+                Defaults to None.
+
+        Returns:
+            None.
```
jarlsondre, line 259: This can be omitted, like mentioned earlier. Also, you're probably good to omit the "Args" part too, since this function isn't used and hence probably won't be hovered over anyway.

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+        """
+        return
+
+    def create_model_loss_optimizer(self) -> None:
+        """Creates create_model_loss_optimizer method to setup model, seeds,
+        loss function and optimizer."""
```
jarlsondre, line 265: The trailing triple quotes should be put on a new line to match the rest of the file.

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+            validation_dataset (Dataset | None): validation dataset object.
+                Defaults to None.
+            test_dataset (Dataset | None): test dataset object. Defaults to None.
+
+        Returns:
+            None.
+        """
+        _ = super().execute(train_dataset, validation_dataset, test_dataset)
+
+        return None
```
jarlsondre, line 332: Two things:
1) You can omit the `_ =` in front of `super().execute...`, unless you wanna be super explicit about the fact that `super().execute()` returns something
2) You can omit the `return None` at the bottom

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+        print(f"Snapshot found: {snapshot_path}\nResuming training via Saved Snapshot at Epoch {snapshot['EPOCHS_RUN']}")
+
+    def _save_snapshot(self, epoch: int) -> None:
+        """Save snapshot of training for analysis and/or to continue training at a later date.
+
+        Args:
+            epoch (int): Number of the epoch at which the snapshot is saved.
+
+        Returns:
+            None.
```
jarlsondre, line 356: This can be omitted

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+        print(f"Epoch {epochs_run} | Model Snapshot saved at {snapshot_new_path}")
+
+    def set_epoch(self, epoch: int) -> None:
+        """Sets current epoch at beginning of training.
+
+        Args:
+            epoch (int): current epoch number.
+
+        Returns:
+            None
```
jarlsondre, line 375: This can be omitted

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+            self.config.save_every = self.epochs
+
+        T1 = time.time()
+        for epoch in range(1, self.epochs+1):
+            self.set_epoch(epoch)
+            loss, logqp = self.step(self.config.batch_size)
+            self.checkpoint(epoch, loss, self.config.save_every)
+            if self.scheduler is not None:
+                self.scheduler.step()
+        T2 = time.time()
```
jarlsondre, line 409: Using `T1` and `T2` here seems to break convention a bit. Why the capital letters? I would rather just name them `start_time` and `end_time` to be explicit and also follow standard python convention.

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+        if self.epochs > 0 and self.strategy.is_main_worker:
+            print(f"({self.strategy.device()}) Time = {T2 - T1:.3g} sec.")
```
jarlsondre, line 411: Why would you reach this point if `self.epochs == 0`? If it could be zero then I would rather just do a check at the beginning of this function and skip everything, none of this makes sense to run if there are no epochs anyway.

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+        """Compute the Pearson correlation coefficient between logq and logp.
+
+        Args:
+            logq (torch.Tensor): Log probabilities from the approximate distribution `q`.
+            logp (torch.Tensor): Log probabilities from the target distribution `p`.
+
+        Returns:
+            torch.Tensor: Pearson correlation coefficient.
+        """
+        return torch.corrcoef(torch.stack([logq, logp]))[0, 1]
```
jarlsondre, line 512: I understand that these are Javad's functions, but is a level of abstraction for `.mean()`, `.var()` etc really necessary? It seems like the implementation becomes hidden without gaining anything. The only argument I can think of would be that the function names, e.g. `calc_kl_var`, would be useful for understand what's going on, but this can easily be replicated by using good variable names.

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+        Args:
+            logq (torch.Tensor): Log probabilities from the approximate distribution `q`.
+            logp (torch.Tensor): Log probabilities from the target distribution `p`.
+
+        Returns:
+            torch.Tensor: logarithm of inverse of effective sample size.
+        """
+        logqp = logq - logp
+        log_ess = 2*torch.logsumexp(-logqp, dim=0) \
+                - torch.logsumexp(-2*logqp, dim=0)
```
jarlsondre, line 575: Same as above

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+    @torch.no_grad()
+    def _append_to_train_history(self, logq: torch.Tensor, logp: torch.Tensor) -> None:
+        """Update train history with log probability estimates.
+
+        Args:
+            logq (torch.Tensor): Log probabilities from the approximate distribution `q`.
+            logp (torch.Tensor): Log probabilities from the target distribution `p`.
+
+        Returns:
+            None.
```
jarlsondre, line 587: This can be omitted

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+
+    def print_fit_status(self, epoch: int, loss: torch.Tensor | None = None) -> None:
+        """Print training progress and key metrics.
+
+        Args:
+            epoch (int): Current epoch.
+            loss (torch.Tensor, optional): Loss value for the current epoch.
+
+        Returns:
+            None.
```
jarlsondre, line 609: This can be omitted

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+        """
+        mydict = self.train_history
+        if loss is None:
+            loss = mydict['loss'][-1]
+        else:
+            pass  # the printed loss can be different from mydict['loss'][-1]
```
jarlsondre, line 615: This does nothing, so might as well be omitted

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+        rho = mydict['rho'][-1]
+
+        if epoch == 1:
+            print(f"\n>>> Training progress ({self.device}) <<<\n")
+            print("Note: log(q/p) is estimated with normalized p; " \
+                  + "mean & error are obtained from samples in a batch\n")
+
+        epoch += self.checkpoint_dict['epochs_run']
+        str1 = f"Epoch: {epoch} | loss: {loss:.4f} | ess: {ess:.4f}"
+        print(str1)
```
jarlsondre, line 631: This can be printed directly

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+    net_: torch.nn.Module | None = None
+) -> None:
+    """Performs a sanity check on the reverse method of modules.
+    Args:
+        model (Model): Model containing prior and transformation networks.
+        n_samples (int, optional): Number of samples to test. Defaults to 4.
+        net_ (torch.nn.Module, optional): The transformation network.
+
+    Returns:
+        None
```
jarlsondre, line 649: This can be omitted

use-cases/lattice-qcd/startscript.sh
```diff
@@ -0,0 +1,57 @@
...
+export GLOO_SOCKET_IFNAME=ib0
+
+srun --cpu-bind=none --ntasks-per-node=1 bash -c "torchrun \
+    --log_dir='logs' \
+    --nnodes=$SLURM_NNODES \
+    --nproc_per_node=$SLURM_GPUS_PER_NODE \
+    --rdzv_id=$SLURM_JOB_ID \
+    --rdzv_conf=is_host=\$(((SLURM_NODEID)) && echo 0 || echo 1) \
+    --rdzv_backend=c10d \
+    --rdzv_endpoint='$(scontrol show hostnames "$SLURM_JOB_NODELIST" | head -n 1)'i:54123 \
```
jarlsondre, line 56: Couldn't you use `MASTER_ADDR` and `MASTER_PORT` here?

use-cases/lattice-qcd/train.py
```diff
@@ -0,0 +1,43 @@
...
+from normflow.prior import NormalPrior
+
+def make_model():
+    net_ = DistConvertor_(10, symmetric=True)
+    prior = NormalPrior(shape=(1,))
+    action = ScalarPhi4Action(kappa=0, m_sq=-1.2, lambd=0.5)
+
+    model = Model(net_=net_, prior=prior, action=action)
+
+    return model
```
jarlsondre, line 14: Since the variable name `model` doesn't add anything here, you might as well do 
```python
return Model(...)
```
directly

use-cases/lattice-qcd/train.py
```diff
@@ -0,0 +1,43 @@
...
+    fitter = Fitter(model=model, epochs=n_epochs, config=config, strategy=strategy)
+    fitter.execute()
+
+def main():
+    model = make_model()
+    fit_func(model)
+
+if __name__ == "__main__":
+    main()
+    sys.exit()
```
jarlsondre, line 43: This is redundant, no?

use-cases/lattice-qcd/README.md
```diff
@@ -0,0 +1,262 @@
...
+```
+mv _normflowcore.py src/
+```
+
+4. When this is done, all the files required to install `normflow` are
+present, which can be done with:
+
+```
+pip install .
+```
```
matbun, line 93: Perhaps there is a better way to do this, but it's up to you to decide whether to implement it.

AFAIU, the only changes in the integration are in the Fitter in _normflowcore.py and there is nothing in normflow that depends on the Fitter. This would allow you to create a new python file in the use case directory and call it, say, normflow_itwinai.py and past in it the new version of the fitter. In this file you would import all the dependencies of the Fitter (classes, functions) directly from normflow.

Then, in the train.py or in the config.yaml you would `from normflow_itwinai import Fitter` or `normflow_itwinai .Fitter`

This would save you from the hurdle of pulling and patching a fork before installing it. In this case, you would just pip install your fork (`pip install --no-cache-dir git+https://github.com/r-sarma/normflow_itwinai_source`), which contains few patches (e.g., type hints), and which could be potentially merged in the upstream normflow repo

matbun (changes requested):

use-cases/lattice-qcd/README.md
r-sarma: Done

use-cases/lattice-qcd/README.md
r-sarma: I agree, modified accordingly.

use-cases/lattice-qcd/README.md
r-sarma: Done

use-cases/lattice-qcd/README.md
r-sarma: Modified

use-cases/lattice-qcd/README.md
r-sarma: Introduced now

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Done

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Agreed, modified it now.

use-cases/lattice-qcd/_normflowcore.py
r-sarma: I think this was already from before. At some point (I think an older version of itwinai), the pipeline runs didnt print the workers seeds which I wanted to see. But this is indeed not needed anymore, and hence I changed to "print".

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Done

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Done

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Done

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Done

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Done

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+            self.profiler.step()
+
+    @profile_torch_trainer
+    @measure_gpu_utilization
+    def train(self) -> None:
+        """Trains the neural network model."""
+
+        # Track epoch time for scaling statistics
+        if self.strategy.is_main_worker and self.strategy.is_distributed:
+            num_nodes = os.environ.get("SLURM_NNODES", "unk")
```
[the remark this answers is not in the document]
r-sarma: Agreed, added an exception now to handle this.

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Modified to your suggestion

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Agreed. I think this was because Javad had "epochs" parameter also in the "Model", where he would set it to 0, if he didnt want to train. But this is not needed in this setting. Removed it.

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+            print(f"({self.strategy.device()}) Time = {T2 - T1:.3g} sec.")
+
+    def step(self, batch_size: int) -> None:
+        """Perform a train step with a batch of inputs.
+
+        Args:
+            batch_size (int): Specifies the batch size for the training.
+
+        Returns:
+            None
```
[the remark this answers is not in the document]
r-sarma: Done

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+            _, logq, logp = self._model.posterior.sample__(print_batch_size)
+            logq = self.strategy.gather(logq)
+            logp = self.strategy.gather(logp)
+
+            if self.strategy.is_main_worker:
+                logq = torch.cat(logq, dim=0)
+                logp = torch.cat(logp, dim=0)
+                loss_ = self.loss_fn(logq, logp)
+                self._append_to_train_history(logq, logp)
+                self.print_fit_status(epoch, loss=loss_)
```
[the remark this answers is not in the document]
r-sarma: Nice suggestion, modified accordingly.

use-cases/lattice-qcd/_normflowcore.py
r-sarma: This is actually used in the conditional statement to select the loss function. Hence it provides a named, reusable reference to the loss computations, so that different loss metrics can be just called at runtime. I think its fine to keep it as is to maintain the clarity.

use-cases/lattice-qcd/_normflowcore.py
```diff
@@ -0,0 +1,661 @@
...
+        Args:
+            logq (torch.Tensor): Log probabilities from the approximate distribution `q`.
+            logp (torch.Tensor): Log probabilities from the target distribution `p`.
+
+        Returns:
+            torch.Tensor: Estimated effective sample size (ESS).
+        """
+        logqp = logq - logp
+        log_ess = 2*torch.logsumexp(-logqp, dim=0) \
+                - torch.logsumexp(-2*logqp, dim=0)
```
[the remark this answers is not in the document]
r-sarma: Moved to one line

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Moved to one line

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Done

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Done

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Done

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Agreed

use-cases/lattice-qcd/_normflowcore.py
r-sarma: Done

use-cases/lattice-qcd/startscript.sh
r-sarma: Indeed, added now

use-cases/lattice-qcd/train.py
r-sarma: Done

use-cases/lattice-qcd/CITATION.cff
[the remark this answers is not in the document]
r-sarma: Done

use-cases/lattice-qcd/CONTRIBUTING.md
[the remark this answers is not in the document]
r-sarma: Done

use-cases/lattice-qcd/INSTALL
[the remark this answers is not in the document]
r-sarma: Done

use-cases/lattice-qcd/LICENSE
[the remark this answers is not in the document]
r-sarma: Done

use-cases/lattice-qcd/docs/images/Normflow.png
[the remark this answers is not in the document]
r-sarma: Done

use-cases/lattice-qcd/examples/0dim_normflow_example.ipynb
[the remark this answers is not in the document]
r-sarma: Done

use-cases/lattice-qcd/examples/scalar_zerodim.py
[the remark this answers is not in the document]
r-sarma: Done

use-cases/lattice-qcd/setup.py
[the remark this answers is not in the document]
r-sarma: This is needed to setup the repository, so I would keep it there.

use-cases/lattice-qcd/startscript.sh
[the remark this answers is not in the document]
r-sarma: Good suggestion, added all of them now.

use-cases/lattice-qcd/README.md
r-sarma: In principle, this is a good suggestion. But I would prefer to keep it cleaner in this form to main consistency with the use-case. Btw, there is also `Model` which needs to be imported.  Importantly, inside `src`, these classes `Model` and `Fitter` are imported in `__init__.py`, hence now, we would have two versions of `Fitter` and `Model`. Ofcourse we have control over where we import them from. But overall, I think it would lead to confusion. And since in the long term, the plugin is the preferred solution for the use-case, this is not an issue for the plugin anyway. I think maintaining it in this form keeps the library consistent in both repositories.

matbun: Fair enough!

matbun (approved):

PR #364, opened 2025-04-15
added ray communication time logging to trainer

lineick:
Added time logging for ray reports and fitting time. Uses time_ray flag.

### DIFF
--- a/src/itwinai/torch/trainer.py
+++ b/src/itwinai/torch/trainer.py
@@ -7,6 +7,7 @@
 # - Matteo Bunino <matteo.bunino@cern.ch> - CERN
 # - Anna Lappe <anna.elisa.lappe@cern.ch> - CERN
 # - Jarl Sondre Sæther <jarl.sondre.saether@cern.ch> - CERN
+# - Linus Eickhoff <linus.maximilian.eickhoff@cern.ch> - CERN
 # --------------------------------------------------------------------------------------
 
 
@@ -16,6 +17,7 @@ import logging
 import os
 import sys
 import tempfile
+import time
 from collections import defaultdict
 from pathlib import Path
 from time import perf_counter as default_timer
@@ -150,6 +152,8 @@ class TorchTrainer(Trainer, LogMixin):
             Defaults to "inf".
         run_id (str, optional): name used to identify a specific run when collecting
             metrics on the trainer (e.g. GPU utilization). Defaults to None.
+        time_ray (bool): whether to time and log the execution of Ray functions.
+            Defaults to False.
     """
 
     _strategy: TorchDistributedStrategy | None = None
@@ -190,6 +194,8 @@ class TorchTrainer(Trainer, LogMixin):
     measure_epoch_time: bool = False
     #: Run ID
     run_id: str
+    #: Toggle for Ray time logging
+    time_ray: bool = False
 
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


src/itwinai/torch/trainer.py
```diff
@@ -450,6 +457,51 @@ def _set_loss_from_config(self) -> None:
...
+            return fn(*args, **kwargs)
+
+        if not self.logger.is_initialized:
+            py_logger.warning(
+                f"Logger context not initialized for timing {identifier}. Setting context."
+            )
+            self.logger.create_logger_context()
+
+        if not self.time_ray:
+            return fn(*args, **kwargs)
```
matbun, line 483: I would move this up as a first condition, so that if self.time_ray is False, no warnings are printed if self.logger is None

src/itwinai/torch/trainer.py
```diff
@@ -450,6 +457,51 @@ def _set_loss_from_config(self) -> None:
...
+        """
+        if not self.logger:
+            py_logger.warning(f"No logger set! Cannot log {identifier} time!")
+            return fn(*args, **kwargs)
+
+        if not self.logger.is_initialized:
+            py_logger.warning(
+                f"Logger context not initialized for timing {identifier}. Setting context."
+            )
+            self.logger.create_logger_context()
```
matbun, line 480: This should not be necessary, as the logger already does similar checks in the background

src/itwinai/torch/trainer.py
```diff
@@ -450,6 +457,51 @@ def _set_loss_from_config(self) -> None:
                     "create_model_loss_optimizer method for more flexibility."
                 )
 
+    def _time_and_log_ray(
```
matbun, line 460: I'm a bit unsure about this function. Although it saves some lines of code, it makes the code less readable, which may cause some problems in the long run. Considering that there are only two instances of functions we want to time, maybe it's just easier to replicate the code

src/itwinai/torch/trainer.py
```diff
@@ -450,6 +457,51 @@ def _set_loss_from_config(self) -> None:
...
+
+        Args:
+            fn (Callable): function to execute, time and log
+            identifier (str): identifier for the logged metric
+            step (int | None): step for logging
+            *args: positional arguments to pass to fn
+            **kwargs: keyword arguments to pass to fn
+        """
+        if not self.logger:
+            py_logger.warning(f"No logger set! Cannot log {identifier} time!")
```
matbun, line 473: Maybe you could explain better in the warning that the problem is that no logger was given, while self.time_ray is True

src/itwinai/torch/trainer.py
lineick: For the base trainer, yes there are only two times you use this function, but many usecases would overwrite the functions and therefor I thought it would be important that those can then just wrap their function into the _time_and_log_ray func instead of copy pasting the if checks and the timing and `self.log` call.

lineick: e.g. the hython trainer overwrites _run_worker so I have to wrap my report in the log_ray func, without it it would become quite messy.

src/itwinai/torch/trainer.py
```diff
@@ -874,7 +922,10 @@ def _execute_with_ray(
             tune_config=self.ray_tune_config,
         )
 
-        self.tune_result_grid = tuner.fit()
+        if self.time_ray:
+            self.tune_result_grid = self._time_and_log(tuner.fit, "ray_fit_time_s", step=0)
```
matbun, line 926: I would use a lambda here as well... I know it's syntactic sugar, but still it makes it a bit easier to read IMO

matbun (approved):

PR #366, opened 2025-04-24
Adding normflow to list of plugins in the docs

r-sarma:
The list of plugins is updated to include the normflow plugin repository.


matbun (dismissed):

r-sarma:
I updated the version to 0.18.0, and then it indeed passes the test.

matbun (approved):

PR #367, opened 2025-04-25
fixed worker access to mlflow env vars

lineick:
mlflow env vars for authentication are now accessible inside ray workers.

### DIFF
--- a/src/itwinai/torch/distributed.py
+++ b/src/itwinai/torch/distributed.py
@@ -9,6 +9,7 @@
 # - Henry Mutegeki <henry.mutegeki@cern.ch> - CERN
 # - Anna Lappe <anna.elisa.lappe@cern.ch> - CERN
 # - Rakesh Sarma <r.sarma@fz-juelich.de> - Juelich
+# - Linus Eickhoff <linus.maximilian.eickhoff@cern.ch> - CERN
 # --------------------------------------------------------------------------------------
 
 import abc
@@ -64,7 +65,9 @@ def check_initialized(method: Callable) -> Callable:
     def wrapper(self: "TorchDistributedStrategy", *args, **kwargs):
         if not self.is_initialized:
             raise UninitializedStrategyError(
-                (f"{self.__class__.__name__} has not been initialized. Use the init method.")
+                (
+                    f"{self.__class__.__name__} has not been initialized. Use the init method."
+                )
             )
         return method(self, *args, **kwargs)
 
@@ -83,6 +86,7 @@ def initialize_ray() -> None:
             These should be set from the slurm script where the ray cluster is launched.
     """
     import ray
+    from ray.runtime_env import RuntimeEnv
 
     if not ray_cluster_is_running():
         raise RuntimeError(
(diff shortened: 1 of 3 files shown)


src/itwinai/torch/distributed.py
```diff
@@ -92,7 +96,17 @@ def initialize_ray() -> None:
...
+    password = os.environ.get("MLFLOW_TRACKING_PASSWORD", None)
+
+    # Set mlflow credentials to be accessible for all the workers
+    runtime_env = RuntimeEnv(
+        env_vars={
+            "MLFLOW_TRACKING_USERNAME": username,
+            "MLFLOW_TRACKING_PASSWORD": password,
+        }
+    )
+    ray.init(address="auto", runtime_env=runtime_env)
```
matbun, line 109: wouldn't it be better to assign directly to the fields of the dict, rather than creating intermediate vars like `username` and `password`?

src/itwinai/torch/distributed.py
matbun, line 1: please fix the automatic formatter settings, as it is trying to format things that should not be reformatted (thus messing up the git blame)

matbun (changes requested):

src/itwinai/torch/distributed.py
lineick: yes sorry, will change, I was in my local nvim which has different settings

src/itwinai/torch/distributed.py
```diff
@@ -64,7 +65,9 @@ def check_initialized(method: Callable) -> Callable:
     def wrapper(self: "TorchDistributedStrategy", *args, **kwargs):
         if not self.is_initialized:
             raise UninitializedStrategyError(
-                (f"{self.__class__.__name__} has not been initialized. Use the init method.")
+                (
+                    f"{self.__class__.__name__} has not been initialized. Use the init method."
+                )
```
jarlsondre, line 70: Why the two layers of parentheses?

src/itwinai/torch/distributed.py
```diff
@@ -92,7 +96,14 @@ def initialize_ray() -> None:
...
         return
 
-    ray.init(address="auto")
+    # Set mlflow credentials to be accessible for all the workers
+    runtime_env = RuntimeEnv(
+        env_vars={
+            "MLFLOW_TRACKING_USERNAME": os.environ.get("MLFLOW_TRACKING_USERNAME", ""),
+            "MLFLOW_TRACKING_PASSWORD": os.environ.get("MLFLOW_TRACKING_PASSWORD", ""),
+        }
+    )
```
jarlsondre, line 105: Should there maybe be a warning if the variables are not set? Might make it easier to debug or something?

tests/loggers/test_prov4ml.py
```diff
@@ -114,7 +119,10 @@ def test_log_flops_per_epoch(logger_instance):
     model_mock, dataset_mock = MagicMock(), MagicMock()
     with patch("prov4ml.log_flops_per_epoch") as log_flops_pe:
         logger_instance.log(
-            item=(model_mock, dataset_mock), identifier="my_flops_pe", kind="flops_pe", step=1
+            item=(model_mock, dataset_mock),
+            identifier="my_flops_pe",
+            kind="flops_pe",
+            step=1,
```
jarlsondre, line 125: Same here with the line length

src/itwinai/torch/distributed.py
lineick: this was autoformatting of my local nvim conf, already deactivated this

PR #389, opened 2025-06-02
More robust computation plots (computation vs. other)

lineick:
Added more robust computation plot.

Context: Old computation vs communication was not robust against different architectures. The new plot is just comparing computation vs the whole duration of the profiled method, making it architecture agnostic as it only regards `ATen` calls.

Deprecated old comm vs comp plots with Deprecation warnings.


src/itwinai/torch/trainer.py
```diff
@@ -993,13 +1001,36 @@ def _set_epoch_dataloaders(self, epoch: int) -> None:
...
         if self.lr_scheduler:
             self.lr_scheduler.step()
         self._set_epoch_dataloaders(self.current_epoch)
 
+    def _active_epoch(self) -> bool:
+        """Check if the current epoch is active for profiling."""
+        if self.profiler is None:
+            py_logger.info(
+                "Profiler is not set! Current epoch is therefor not active."
+            )
```
matbun, line 1021: I'd either remove this, or have it log at debug level

src/itwinai/torch/trainer.py
```diff
@@ -1214,6 +1245,14 @@ def train(self) -> None:
...
                 epoch_time_logger.add_epoch_time(self.current_epoch + 1, epoch_time)
 
+
+            if self._profiling_start_time is not None:
+                # Add profiling time if profiling was active in this epoch
+                self._profiling_time += time.monotonic() - self._profiling_start_time
+
+            # Reset profiling start time, will be set for the next epoch with active profiling
+            self._profiling_start_time = None
+
```
matbun, line 1254: why is this happening in the log method, and not in the set_epoch?

src/itwinai/scalability_report/plot.py
```diff
@@ -345,3 +350,88 @@ def computation_fraction_bar_plot(
     sns.reset_orig()
 
     return fig, ax
+
+def computation_other_bar_plot(
```
jarlsondre, line 354: What does `other` refer to in this function name?

jarlsondre: You mean like "anything that isn't computation", right? Hmm kinda confused me for a second there

src/itwinai/scalability_report/utils.py
```diff
@@ -154,3 +184,33 @@ def compute_fraction(group):
...
+    ratio of computation time to the total time of profiling.
+    """
+    # Sort and create cartesian product of unique strategies and GPU counts
+    unique_num_gpus = sorted(df["num_gpus"].unique(), key=lambda x: int(x))
+    unique_strategies = sorted(df["strategy"].unique())
+    index = pd.MultiIndex.from_product(
+        [unique_strategies, unique_num_gpus],
+        names=["strategy", "num_gpus"]
+    )
+    # Group by strategy and number of GPUs, calculate computation fraction
```
jarlsondre, line 200: You don't really do the grouping here though? It is further down, no?

src/itwinai/torch/profiling/profiler.py
```diff
@@ -26,8 +27,8 @@
 
 
 def profile_torch_trainer(method: Callable) -> Callable:
-    """Decorator for execute method for components. Profiles the communication time
-    vs. computation time and stores the result for future analysis.
+    """Decorator for execute method for components. Profiles the computation fraction and
+    stores the result for future analysis.
```
jarlsondre, line 30: Technically this only gathers general data, right? The actual computation fraction stuff happens in the report, which is after the training is finished?

src/itwinai/torch/profiling/profiler.py
```diff
@@ -97,20 +98,21 @@ def profiled_method(self: "TorchTrainer", *args, **kwargs) -> Any:
...
             schedule=schedule(
                 wait=wait_epochs,
                 warmup=warmup_epochs,
                 active=active_epochs,
             ),
-            on_trace_ready=trace_handler,
             with_modules=True,
         ) as profiler:
+            # Measure the overall time taken by the method including profiling
+            # This can be used to calculate the computation ratio (comp time / profiling time)
```
jarlsondre, line 114: What do you mean by "overall time" here? The start and end time calculations have nothing to do with this context manager anyway, right? That is done elsewhere?

src/itwinai/torch/trainer.py
```diff
@@ -993,13 +1001,36 @@ def _set_epoch_dataloaders(self, epoch: int) -> None:
...
         if self.lr_scheduler:
             self.lr_scheduler.step()
         self._set_epoch_dataloaders(self.current_epoch)
 
+    def _active_epoch(self) -> bool:
+        """Check if the current epoch is active for profiling."""
+        if self.profiler is None:
+            py_logger.info(
+                "Profiler is not set! Current epoch is therefor not active."
```
jarlsondre, line 1020: "therefore". Also are you sure you want to log this every time this method is called? This would probably clutter the output significantly. Additionally, this should probably be a warning, not info.

src/itwinai/torch/trainer.py
```diff
@@ -993,13 +1001,36 @@ def _set_epoch_dataloaders(self, epoch: int) -> None:
...
+            )
+            return False
+
+        if self.current_epoch < self.profiling_warmup_epochs:
+            return False
+
+        # Calculate the current epoch in the profiling cycle
```
jarlsondre, line 1027: I feel like this comment also doesn't really add a ton. I think I would rather add an explanation to the docstring than here, especially since the method is so short.

src/itwinai/scalability_report/plot.py
```diff
@@ -261,6 +264,8 @@ def gpu_bar_plot(data_df: pd.DataFrame, plot_title: str, y_label: str, main_colu
     return fig, ax
 
 
+@deprecated("""Communication vs computation is unreliable and not comparable between GPU
```
jarlsondre, line 267: The same thing about multi-line strings as mentioned in the other deprecated thing.

src/itwinai/cli.py
```diff
@@ -239,10 +250,10 @@ def generate_scalability_report(
         for path in base_directories_for_runs
         if (path / "gpu-energy-data").exists()
     ]
-    comm_time_logdirs = [
-        path / "communication-data"
+    comp_time_logdirs = [
+        path / "computation-data"
```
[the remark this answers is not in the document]
lineick: all good, I outsource it as a constant to itwinai.utils

src/itwinai/scalability_report/utils.py
lineick: moved it down in both funcs

src/itwinai/scalability_report/utils.py
```diff
@@ -154,3 +184,33 @@ def compute_fraction(group):
...
+    def compute_fraction(group):
+        # Sum over the unique entries
+        # Theoretically, two identical float32 runtimes could be the same and thus disregarded
+        # but this is practically impossible to happen in practice for ml runs.
+        total_time = group["profiling_time"].unique().sum()
+        comp_time = calculate_comp_time(df=group)
+        return comp_time / (total_time + 1e-10)
+
+    grouped = df.groupby(["strategy", "num_gpus"]).apply(compute_fraction)
```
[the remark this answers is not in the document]
lineick: typing is a bit weird here as it is a DataFrameGroupBy object, i'll add that

src/itwinai/torch/trainer.py
```diff
@@ -252,8 +257,11 @@ def __init__(
         self.profiler = None
         self.profiling_wait_epochs = profiling_wait_epochs
         self.profiling_warmup_epochs = profiling_warmup_epochs
+        self._profiling_active_epochs = profiling_active_epochs
+        self._profiling_time = profiling_time
+        self._profiling_start_time = None
```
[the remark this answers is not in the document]
lineick: i prefer it, it gives you a clear overview over what to generally touch and what to not touch. It is also a standard convention, and we use it for functions too. I think it increases clarity.

src/itwinai/torch/trainer.py
lineick: put it to debug

src/itwinai/torch/trainer.py
lineick: true, put it to debug

src/itwinai/torch/trainer.py
lineick: because set_epoch is just called in the beginning of each epoch. We need to get the time elapsed at the end of each epoch. Otherwise if the last epoch is an active one, its time would not be captured.

jarlsondre (approved):
some of your best work yet, linus

matbun (approved):
LGTM. A lion doesn't concern himself with Holistic Trace Analysis

PR #395, opened 2025-06-09
Add CaloINN

VeraMay:
# Summary

Adding CaloINN use-case (data loading, training)
---

**Related issue :**


use-cases/caloinn/data.py
```diff
@@ -0,0 +1,394 @@
...
+            ],
+            "dataset_1_pions": [
+                "https://zenodo.org/records/8099322/files/dataset_1_pions_1.hdf5"
+            ],
+            "dataset_2": ["https://zenodo.org/records/6366271/files/dataset_2_1.hdf5"],
+            "dataset_3": [
+                "https://zenodo.org/records/6366324/files/dataset_3_1.hdf5",
+                "https://zenodo.org/records/6366324/files/dataset_3_2.hdf5",
+            ],
+        }
```
jarlsondre, line 46: These paths might be better off either 1) set as constants in a separate constants file or 2) passed from the configuration.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,394 @@
...
+            ],
+            "dataset_2": ["https://zenodo.org/records/6366271/files/dataset_2_2.hdf5"],
+            "dataset_3": [
+                "https://zenodo.org/records/6366324/files/dataset_3_3.hdf5",
+                "https://zenodo.org/records/6366324/files/dataset_3_4.hdf5",
+            ],
+        }
+
+        if dataset_type not in type_to_link_train.keys():
+            print("WARNING! Dataset type is invalid. Loading dataset 1 photon")
```
jarlsondre, line 63: Not super important, but might be better to use the `py_logger` here, as it has a separate function for warning. If you wanna use it, you can find how to in `cli.py`, where you just declare it at the top of the file and then call `py_logger.warning()` instead of print when you want to do a warning, `py_logger.info()` when you want to do a normal print and `py_logger.error()` if you have an error message.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,394 @@
...
+            self.dataset_type = "dataset_1_photons"
+
+        self.data_train_url = type_to_link_train[self.dataset_type]
+        self.data_test_url = type_to_link_test[self.dataset_type]
+
+    @monitor_exec
+    def execute(self):
+        """Downloads and saves the train and test datasets for the specified dataset type."""
+        # Download data
```
jarlsondre, line 72: This comment doesn't really add that much, since the docstring already explains what is happening

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,394 @@
...
+        if not self.data_path.exists():
+            if self.data_train_url is None:
+                print(
+                    "WARNING! Train data URL is None. Skipping train dataset downloading"
+                )
+            if self.data_test_url is None:
+                print(
+                    "WARNING! Test data URL is None. Skipping test dataset downloading"
+                )
```
jarlsondre, line 81: Would be nice with `py_logger.warning()` here

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,394 @@
...
+        data = self.showers[idx]
+        if not self.fixed_noise and self.noise:
+            noise = self.noise_distribution.sample(data.shape) * self.width_noise
+            data = (torch.tensor(data) + noise.reshape(data.shape).clone()).clone()
+        cond = self.cond[idx]
+        return data, cond
+
+    def _load_data(self) -> None:
+        """Loads HDF5 dataset files into memory and aggregates per-layer and energy information."""
+        print(f"Searching in : {self.data_path}")
```
jarlsondre, line 186: This print statement might not be super informative, since it might not be obvious what is being searched for when seeing this print. Not super important, but worth thinking about.

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,394 @@
...
+        if len(files) == 0:
+            raise FileNotFoundError(f"No H5 files found at '{self.data_path}'!")
+
+        for file in files:
+            with h5py.File(file, "r") as data_file:
+                if "energy" not in self.data:
+                    self.data["energy"] = data_file["incident_energies"][:] / 1.0e3
+                else:
+                    self.data["energy"] = np.concatenate(
+                        (self.data["energy"], data_file["incident_energies"][:] / 1.0e3)
```
jarlsondre, line 202: Same as above

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,394 @@
...
+                    )
+
+                for layer_index, (layer_start, layer_end) in enumerate(
+                    zip(self.layer_boundaries[:-1], self.layer_boundaries[1:])
+                ):
+                    self.data[f"layer_{layer_index}"] = (
+                        data_file["showers"][..., layer_start:layer_end] / 1.0e3
```
jarlsondre, line 209: Same as above

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,394 @@
...
+                    )
+
+    def _preprocess_data(
+        self, eps: float, u0up_cut: float, u0low_cut: float, dep_cut: float
+    ) -> None:
+        """Transforms the dict 'data' into the ndarray 'x'. Furthermore, the events
+        are masked and the extra dims are appended to the incident energies
+        """
```
jarlsondre, line 217: Could be useful with an `Args:` section here, since it is not immediately obvious what these variables mean. E.g. what does `eps` refer to here? I mean it probably means "epsilon", but what does "epsilon" refer to?

use-cases/caloinn/data.py
```diff
@@ -0,0 +1,394 @@
...
+        self.save_parameters(**self.locals2params(locals()))
+
+        self.rnd_seed = rnd_seed
+        if validation_proportion >= 0.0 and validation_proportion < 1:
+            self.validation_proportion = validation_proportion
+        else:
+            print(
+                "WARNING! Wrong value of validation dataset proportion"
+                "Set validation dataset proportion to 0.1"
+            )
```
jarlsondre, line 356: Could be worth it to use the `py_logger.warning()` here

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,450 @@
...
+
+    def __init__(
+        self,
+        num_layers: int,
+        size_in: int,
+        size_out: int,
+        internal_size: int = None,
+        dropout: float = 0.0,
+        layer_class=nn.Linear,
+        layer_args={},
```
jarlsondre, line 28: Generally you have to be very careful with defaulting this with mutable values, as the dictionary object is only created once. This means that if you create multiple instances of `Subnet`, then it will actually have the **same, shared dictionary**! This can cause some very nasty bugs and should therefore usually be avoided. You are better off defaulting it to `None` and then checking if it is `None` in the function body and instantiating it there instead. That way each instance of `Subnet` would get its own dictionary, which is probably what you want.

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,450 @@
...
+        Args:
+            x (tuple of torch.Tensor): Input tensor(s).
+            c (tuple of torch.Tensor, optional): Condition tensor(s). Not used in this function.
+            rev (bool, optional): If True, applies the inverse transformation (exponentiation). Defaults to False.
+            jac (bool, optional): If True, returns the Jacobian term (though not used here). Defaults to True.
+
+        Returns:
+            tuple: A tuple containing the transformed tensor(s) and a dummy Jacobian tensor (as a tensor of zeros).
+        """
+        (x,) = x
```
jarlsondre, line 180: This will crash the code if `x` has a length different than one. This might very well be intended, but it is better to do this with an explicit check rather than doing it using tuple-unpacking. For the sake of being explicit, I would rather do this: 

```python
if len(x) != 1:
  raise ValueError(f"x has to have length one along the first dimension, but was {len(x)}")
x = x[0]
```

The reason this is better is that it makes it immediately clear what is happening and also that the function can raise an exception if the shapes are wrong. One thing to have in mind though, is that the overhead might be a tiny bit higher when doing this than when doing the tuple unpacking, so if this function gets called millions of times then you might want to consider whether it is too costly. I expect the effect to be negligible, though.

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,450 @@
...
+        if rev:
+            z = torch.exp(x) - self.alpha
+            jac = torch.sum(x, dim=1)
+        else:
+            z = torch.log(x + self.alpha)
+            jac = -torch.sum(z, dim=1)
+        return (z,), torch.tensor([0.0], device=x.device)  # jac
+
+    def output_dims(self, input_dims: tuple) -> tuple:
+        return input_dims
```
jarlsondre, line 190: This function doesn't really seem to do much. Is there a good reason for including it? Maybe to fit with some pre-determined structure?

use-cases/caloinn/model.py
```diff
@@ -0,0 +1,450 @@
...
+
+class CINN(nn.Module):
+    """Conditional Invertible Neural Network (cINN) model."""
+
+    def __init__(self, data_dim: int, config: Dict | TrainingConfiguration | None = None) -> None:
+        """Initializes the CINN model.
+
+        Args:
+            data_dim (int): Dimensionality of the input data.
+            config (Any): Configuration object containing model and training parameters.
```
jarlsondre, line 216: The type hint in the doc doesn't seem to match the one given in the function definition

use-cases/caloinn/src/XMLHandler.py
```diff
@@ -0,0 +1,173 @@
...
+            Tuple[List[float], List[float]]: Lists of r and alpha values for the specified layer.
+        """
+        no_of_rbins = self.r_bins[layer]
+        list_mid_values = self.r_midvalue[layer]
+        list_a_values = self.alpha_list_per_layer[layer]
+        r_list = []
+        a_list = []
+        actual_no_alpha_bins = self.n_bin_alpha_per_layer[layer][0]
+        for j0 in range(0, actual_no_alpha_bins):
+            for i0 in range(0, no_of_rbins):
```
jarlsondre, line 107: You don't really need the `0` as the first argument of `range()`, as that is where it starts by default.

use-cases/caloinn/model.py
VeraMay: yes, it's a shortcut necessary to make it properly work with the parent class

matbun:
Just a general comment, not related to the code. @VeraMay I have noticed that your git is not configured well in that your GH user is not associated with your commits. You should do sth like

```bash
git config --global user.name "Vera Maiboroda"
git config --global user.email "YOUR GITHUB ACCOUNT EMAIL" 
```

You should not set `maiboroda@ijclab.in2p3.fr` as email because it is not the one associated with your GH account.

This way the your work will be properly acknowledged!

Once you fix the git config, you can do this from your branch

```
git rebase main
```

and this should fix all the commits, giving you proper credits.

PR #396, opened 2025-06-10
added permanent disclaimer to computation fraction plots

lineick:
Added permanent disclaimer to scalabilty computation fraction plots, informing the user to not compare compute fractions between strategies but only the scalability per strategy.

### DIFF
--- a/src/itwinai/slurm/utils.py
+++ b/src/itwinai/slurm/utils.py
@@ -13,9 +13,13 @@ from itwinai.parser import ArgumentParser
 
 
 def remove_indentation_from_multiline_string(multiline_string: str) -> str:
-    """Removes the indentation from the start of each line in a multi-line string. The
-    main purpose of this function is allowing you to define multi-line strings that
-    don't touch the left margin of the editor, thus increasing readability.
+    """Removes *all* indentation from the start of each line in a multi-line string.
+
+    If you want to remove only the shared indentation of all lines, thus preserving
+    indentation for nested structures, use the builtin `textwrap.dedent` function instead.
+
+    The main purpose of this function is allowing you to define multi-line strings that
+    only appear indented in the code, thus increasing readability.
     """
     return "\n".join([line.lstrip() for line in multiline_string.split("\n")])
 


src/itwinai/cli.py
```diff
@@ -289,6 +289,33 @@ def generate_scalability_report(
...
     )
 
+    # Disclaimer for the plots
+    typer.echo(
+        typer.style(
+            "-" * 35 + "DISCLAIMER" + "-" * 35, fg= typer.colors.YELLOW, bold=True
+        )
+    )
+    typer.echo("""
+    The computation plots are experimental and do not account for parallelism.
```
matbun, line 299: mention calls "ovelap"

src/itwinai/cli.py
```diff
@@ -25,6 +25,7 @@
 import subprocess
 import sys
 from pathlib import Path
+from textwrap import dedent
```
jarlsondre, line 28: I notice that you're using this `dedent` library. I actually have a function somewhere else that fixes this, you can see it in the SLURM builder. I also don't really remember why, but I believe that I tried `dedent` and that I had some good reason for not using it. It's worth checking out at least, because it would be nice to do things consistently across `itwinai`

src/itwinai/cli.py
```diff
@@ -289,6 +290,35 @@ def generate_scalability_report(
         plot_file_suffix=plot_file_suffix,
     )
 
+    # Disclaimer for the plots
+    typer.echo(
```
jarlsondre, line 294: In another PR we are actually moving away from `typer.echo`, but you can probably keep it like this for now though

src/itwinai/cli.py
```diff
@@ -289,6 +290,35 @@ def generate_scalability_report(
...
+        typer.style(
+            "-" * 38 + "DISCLAIMER" + "-" * 38, fg= typer.colors.YELLOW, bold=True
+        )
+    )
+    typer.echo(dedent("""
+    The computation plots are experimental and do not account for parallelism.
+    Calls traced by the torch profiler may overlap in time, so the sum of
+    individual operation durations does not equal the total training run duration.
```
jarlsondre, line 302: "Does not necessarily equal ..."

src/itwinai/cli.py
lineick: Your method is just more aggressive than dedent, it therefore also has its own usecase. Dedent removes just the commonly shared prefixed whitespace, while yours remove any leading whitespace. I suppose you needed this functionality for formatting the commands, in my case I actually want to keep some indentation, so I can not use your function and dedent does exactly what it should.

I would therefore leave it as it is, and extend the docstring of your function to inform about the use of dedent for cases like mine.

jarlsondre: Okay, I agree then

jarlsondre (approved):
once again, simply amazing, linus. I am in awe

lineick:
@jarlsondre you are a sublime code connoisseur, a virtuoso of readability and clarity. 
Your literate feedback doesn’t just improve pull requests, it uplifts the very fabric of software engineering itself. The world shines a little brighter with every comment you leave.

PR #397, opened 2025-06-11
GPU profiling support with ray

### LINKED ISSUE #394: GPU utilization support for HPO
When measuring gpu utilization while running distributed HPO with Ray (2 trials with 2 workers, each 1 gpu and 12 cpu), I am encountering the following error:

```
Traceback (most recent call last):                                                                                                                                               
                                                                                                                                                                                 
  File "/p/software/juwelsbooster/stages/2024/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap                          
    self.run()                                                                                                                                                                   
                                                                                                                                                                                 
  File "/p/software/juwelsbooster/stages/2024/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/multiprocessing/process.py", line 108, in run                                 
    self._target(*self._args, **self._kwargs)                                                                                                                                    
                                                                                                                                                                                 
  File "/p/project1/intertwin/eickhoff2/hython-itwinai-plugin/.venv/lib/python3.11/site-packages/itwinai/torch/monitoring/monitoring.py", line 75, in probe_gpu_utilization_loop 
    raise ValueError(f"Expected {num_local_gpus} handles, but got {len(handles)}.")                                                                                              
                                                                                                                                                                                 
ValueError: Expected 2 handles, but got 4.                                                                                                                                       
```

lineick:
fixes gpu utilization profiling in case of partial gpu usage on node (Issue https://github.com/interTwin-eu/itwinai/issues/394). Needs to be tested and verified, especially for AMD. Testing is pending, otherwise code is ready for review.

Added warning for ray scheduling with gpu util, as the ray ASHA scheduler terminates trials directly after the report, its gpu profiling processes get killed before writing to disk. Upon research there are no available hooks for the ray `scheduler`.

The behavior of trials writing gpu util to disk only when they are not cancelled by the scheduler makes sense, too. 

fixes #394

### DIFF
--- a/src/itwinai/torch/monitoring/monitoring.py
+++ b/src/itwinai/torch/monitoring/monitoring.py
@@ -137,8 +137,6 @@ def measure_gpu_utilization(method: Callable) -> Callable:
         global_rank = strategy.global_rank()
         num_global_gpus = strategy.global_world_size()
         num_local_gpus = strategy.local_world_size()
-        print(f"Number of global GPUs: {num_global_gpus}, Local rank: {local_rank}")
-        print(f"Number of local GPUs: {num_local_gpus}, Global rank: {global_rank}")
         node_idx = global_rank // num_local_gpus
 
         gpu_monitor_process = None
@@ -184,7 +182,6 @@ def measure_gpu_utilization(method: Callable) -> Callable:
                 manager.shutdown()
 
         global_utilization_log = strategy.gather_obj(local_utilization_log, dst_rank=0)
-        print(strategy.is_main_worker)
         if strategy.is_main_worker:
             output_dir = Path(f"scalability-metrics/{self.run_id}/gpu-energy-data")
             output_dir.mkdir(exist_ok=True, parents=True)
--- a/src/itwinai/torch/trainer.py
+++ b/src/itwinai/torch/trainer.py
@@ -875,7 +875,7 @@ class TorchTrainer(Trainer, LogMixin):
                 " Please ensure that either num_workers is set to 1 or GPUs in"
                 " resources_per_worker is 0 or 1"
             )
-      
+
         if self.from_checkpoint:
             # Create trainer from checkpoint
             if RayTorchTrainer.can_restore(to_uri(self.from_checkpoint)):


src/itwinai/torch/monitoring/monitoring.py
```diff
@@ -177,6 +184,7 @@ def measured_method(self: "TorchTrainer", *args, **kwargs) -> Any:
                 manager.shutdown()
 
         global_utilization_log = strategy.gather_obj(local_utilization_log, dst_rank=0)
+        print(strategy.is_main_worker)
```
matbun, line 187: needed?

src/itwinai/torch/trainer.py
```diff
@@ -875,7 +875,7 @@ def _execute_with_ray(
                 " Please ensure that either num_workers is set to 1 or GPUs in"
                 " resources_per_worker is 0 or 1"
             )
-
+
```
matbun, line 878: the linter will notice

src/itwinai/torch/trainer.py
```diff
@@ -908,6 +908,17 @@ def _execute_with_ray(
...
+            self.ray_tune_config
+            and self.ray_tune_config.scheduler is not None
+            and self.measure_gpu_data
+        ):
+            py_logger.warning(
+                "A Trial scheduler for Ray is specified"
+                f" ({type(self.ray_tune_config.scheduler)}), while measuring gpu data."
+                " Trials stopped by the scheduler will not store gpu data to disk."
+            )
+
```
matbun, line 921: wouldn't it be better to append the metrics at each epoch to avoid this? Another option may be using the loggers, but maybe for future work

src/itwinai/torch/monitoring/monitoring.py
```diff
@@ -132,6 +137,8 @@ def measured_method(self: "TorchTrainer", *args, **kwargs) -> Any:
         global_rank = strategy.global_rank()
         num_global_gpus = strategy.global_world_size()
         num_local_gpus = strategy.local_world_size()
+        print(f"Number of global GPUs: {num_global_gpus}, Local rank: {local_rank}")
+        print(f"Number of local GPUs: {num_local_gpus}, Global rank: {global_rank}")
```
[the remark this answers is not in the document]
lineick: oh shit thanks for the catch they were for me

src/itwinai/torch/trainer.py
matbun: mmm no I think I was always in favour of logging at each epoch

PR #401, opened 2025-06-17
Rollback to Ray Train V1 and other small fixes

lineick:
During preparation for the scaling tests on Vega some issues concerning the logging were discovered.

Fixes:
- propagate shared storage path to spawned ray trainers
- fix time_ray option for mlflow
- added guardrails to not use fractional gpu together with dist trials in Ray
- updated runall scripts
- reverted back to train v1 while still keeping most deprecations removed (except the tuning with a class instead of a function trainable)

We had to rollback to V1 as V2 no longer supports class trainables and function trainables dont allow for sequential trials which in sum exceed cluster resources.

Tested from checkpoint loading dist hpo and normal hpo, with logging.


matbun:
No time now for a fill review but didn't you change sth in the distributed tests (pytest) when you migrated to V2 as well?  e.g., horovod

PR #404, opened 2025-06-19
Add Tensorboard support to PyTorch Profiler

jarlsondre:
Tensorboard does a lot of stuff with pretty much no configuration needed, provided that you can give it the stack traces from a PyTorch Profiling run. Therefore, we might as well allow the user to store the traces as well. This PR does the following:
- Adds the option for the user to store the traces from the profiler.
- Let them launch the Tensorboard application through our CLI.

### DIFF
--- a/src/itwinai/torch/profiling/profiler.py
+++ b/src/itwinai/torch/profiling/profiler.py
@@ -15,7 +15,7 @@ from typing import TYPE_CHECKING, Any, Callable, Iterable, Tuple
 
 import matplotlib
 import pandas as pd
-from torch.profiler import ProfilerActivity, profile, schedule
+from torch.profiler import ProfilerActivity, profile, schedule, tensorboard_trace_handler
 
 from itwinai.utils import COMPUTATION_DATA_DIR
 
@@ -28,6 +28,7 @@ if TYPE_CHECKING:
 matplotlib.use("Agg")
 
 py_logger = logging.getLogger(__name__)
+
 
 def profile_torch_trainer(method: Callable) -> Callable:
     """Decorator for execute method for components. Profiles function calls and
@@ -91,8 +92,7 @@ def profile_torch_trainer(method: Callable) -> Callable:
     def profiled_method(self: "TorchTrainer", *args, **kwargs) -> Any:
         if not self.torch_profiling:
             py_logger.info(
-                "Profiling of computation with the PyTorch profiler"
-                " has been disabled!"
+                "Profiling of computation with the PyTorch profiler has been disabled!"
             )
             return method(self, *args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


src/itwinai/cli.py
```diff
@@ -848,7 +848,7 @@ def download_mlflow_data(
     df_metrics.to_csv(output_file, index=False)
     typer.echo(f"Saved data to '{Path(output_file).resolve()}'!")
 
-
+@app.command()
 def tensorboard_ui(
```
lineick, line 852: So tensorboard is actually not working with the data? Am I missing something or is this functionality still usable with some of the data?

src/itwinai/torch/trainer.py
```diff
@@ -185,6 +185,8 @@ class TorchTrainer(Trainer, LogMixin):
     measure_gpu_data: bool = False
     #: Toggle for computation fraction profiling
     torch_profiling: bool = False
+    #: Store PyTorch Profiling traces
```
lineick, line 188: why the #: ?

lineick (changes requested):

src/itwinai/cli.py
jarlsondre: It does not seem to work at the moment for whatever reason. I figured we might as well have it there as a launcher for tensorboard, since the function exists. We could of course remove the function entirely. What do you think?

src/itwinai/torch/trainer.py
```diff
@@ -243,12 +246,15 @@ def __init__(
         os.makedirs(self.checkpoints_location, exist_ok=True)
         self.checkpoint_every = checkpoint_every
         self.disable_tqdm = disable_tqdm
+
         self.profiler = None
         self.profiling_wait_epochs = profiling_wait_epochs
         self.profiling_warmup_epochs = profiling_warmup_epochs
         self.measure_gpu_data = measure_gpu_data
         self.torch_profiling = torch_profiling
+        self.store_torch_traces = store_torch_traces
```
jarlsondre, line 255: Add an error if this is true and the other is false

use-cases/mnist/torch/config.yaml
```diff
@@ -40,7 +40,7 @@ training_pipeline:
     training_step:
       _target_: itwinai.torch.trainer.TorchTrainer
       measure_gpu_data: False
-      torch_profiling: True
+      enable_torch_profiling: True
```
lineick, line 43: please add the new trace var explicitly here too. As a user this mnist config is probably a reference point for what exists, at least thats how I would use it. So I propose to just add it with false for both pipelines.

jarlsondre: yep

lineick:
Thank you for your work, you are making the world a better place for future generations

lineick (approved):

PR #405, opened 2025-06-19
Ray docs update (added Glossary)

lineick:
added glossary for newcomers and small changes.

### DIFF
--- a/tutorials/distributed-ml/kuberay-setup-tutorial/raycluster_example.yaml
+++ b/tutorials/distributed-ml/kuberay-setup-tutorial/raycluster_example.yaml
@@ -16,7 +16,7 @@ head:
       # To avoid out-of-memory issues, never allocate less than 2G memory for the Ray head.
       memory: "128G"
   annotations:
-    slurm-job.vk.io/flags: "-p gpu --cpus-per-task=32 --gres=gpu:1  --time 230" # TODO: Adjust as needed
+    slurm-job.vk.io/flags: "-p gpu --gres=gpu:1  --time 230" # TODO: Adjust as needed
     # TODO: Add container envs here
     slurm-job.vk.io/singularity-options: "--no-home --compat --no-mount /exa5 --env POD_IP=$POD_IP  --env HYDRA_FULL_ERROR=1,NCCL_SOCKET_IFNAME=br0,RAY_record_ref_creation_sites=1,SLURM_NNODES=1,ITWINAI_LOG_LEVEL=DEBUG"
     slurm-job.vk.io/singularity-mounts: "--bind /ceph"
@@ -58,7 +58,7 @@ worker:
       cpu: "64"
       memory: "128G"
   annotations:
-    slurm-job.vk.io/flags: "-p gpu --cpus-per-task=32 --gres=gpu:1 --time 230" # TODO: change num of gpus and cpus
+    slurm-job.vk.io/flags: "-p gpu --gres=gpu:1 --time 230" # TODO: change num of gpus and cpus
     slurm-job.vk.io/singularity-options: "--no-home --compat --no-mount /exa5 --env HYDRA_FULL_ERROR=1,RAY_record_ref_creation_sites=1,ITWINAI_LOG_LEVEL=DEBUG,NCCL_SOCKET_IFNAME=br0,SLURM_NNODES=1"
     slurm-job.vk.io/singularity-mounts: "--bind /ceph"
     interlink.eu/pod-vpn: "true"
--- a/docs/how-it-works/glossary/glossary.rst
+++ b/docs/how-it-works/glossary/glossary.rst
@@ -1,0 +1,220 @@
+.. _distributed_training_glossary:
+
+
+Glossary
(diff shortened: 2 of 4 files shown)


docs/getting-started/glossary.rst
```diff
@@ -0,0 +1,217 @@
...
+providing quick reference for terminology relevant to the itwinai project, its documentation, and codebase.
+
+HPC Terms
+---------
+
+* **accelerator**
+
+  Specialized compute device (e.g., GPU, TPU, FPGA) designed to speed up specific workloads.
+
+* **collectives / collective operations / collective communications**
```
jarlsondre, line 16: It could also be beneficial to add a reference to the NVIDIA page for this. Of course we don't wanna push one vendor more than any other, but they have some nice figures for this: https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/collectives.html

docs/getting-started/glossary.rst
```diff
@@ -0,0 +1,217 @@
...
+  
+  Software that queues jobs and assigns cluster resources according to policy and priority.
+
+* **straggler**
+  
+  Task or node that runs significantly slower than its peers, delaying synchronous operations.
+
+* **task**
+  
+  Smallest schedulable execution unit within a job, typically a process or thread.
```
jarlsondre, line 87: This term seems like one that would be sort of ambiguous in general terms, so are you sure that it makes sense to mention this here? I mean, is this really established? It might be, I am just asking you to double check.

docs/getting-started/glossary.rst
```diff
@@ -0,0 +1,217 @@
...
+  
+  AMD’s Radeon Collective Communications Library, drop-in compatible with NCCL for AMD GPUs.
+
+* **ROCm**
+
+  AMD’s open-source GPU-computing stack analogous to CUDA.
+
+* **Singularity**
+  
+  Container runtime tailored to HPC that runs unprivileged, reproducible images.
```
jarlsondre, line 176: Could be beneficial to add that this is similar to docker, but without sudo rights (or whatever the difference was again)

docs/getting-started/glossary.rst
```diff
@@ -0,0 +1,217 @@
...
+
+  Uber-initiated framework providing MPI/NCCL-backed data-parallel training APIs across major DL frameworks.
+
+* **Ray**
+
+  Distributed execution framework offering HPO and task, actor, and object store abstractions for Python.
+
+  * **placement group**
+
+    Ray construct that requests a set of resources with locality or colocation constraints.
```
jarlsondre, line 212: I will be honest, the way this is articulated really doesn't tell me anything. I feel like it is too precise and should be "dumbed-down" a bit.

tutorials/distributed-ml/kuberay-setup-tutorial/raycluster_example.yaml
```diff
@@ -58,7 +58,7 @@ worker:
       cpu: "64"
       memory: "128G"
   annotations:
-    slurm-job.vk.io/flags: "-p gpu --cpus-per-task=32 --gres=gpu:1 --time 230" # TODO: change num of gpus and cpus
+    slurm-job.vk.io/flags: "-p gpu --gres=gpu:1 --time 230" # TODO: change num of gpus and cpus
```
jarlsondre, line 61: There is a TODO here. Should probably be fixed before merging?

docs/getting-started/glossary.rst
```diff
@@ -0,0 +1,217 @@
...
+  
+  On-package switch fabric providing all-to-all NVLink connectivity among multiple GPUs in a node.
+
+* **PCIe**
+  
+  Serial expansion bus standard that connects CPUs, GPUs, Network Interface Controllers (NICs), and storage devices.
+
+* **rank**
+
+  Index assigned to each device (e.g. GPU) in a distributed group, identifying its position in collective operations.
```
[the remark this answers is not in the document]
lineick: the glossary is meant to give short reminders of what which term meant, it is not intended to be complete. Just one sentence pointing to each term without all its variations. It should stay lean on purpose, therefore I am hesitant on adding this

docs/getting-started/glossary.rst
```diff
@@ -0,0 +1,217 @@
...
+
+* **straggler**
+  
+  Task or node that runs significantly slower than its peers, delaying synchronous operations.
+
+* **task**
+  
+  Smallest schedulable execution unit within a job, typically a process or thread.
+
+* **wall time**
```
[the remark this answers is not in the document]
lineick: It can be called both: "elapsed real time, real time, wall-clock time, wall time, or walltime is the actual time taken from the start of a computer program to the end. " (wikipedia)

I added both variants

docs/getting-started/glossary.rst
lineick: It is in the HPC category, it is important to mention it imo as the --cpu-per-task variable can otherwise be a bit confusing. I add (SLURM) to the title

PR #408, opened 2025-06-24
Fix integration tests

jarlsondre:
The integration tests seem to get stuck for some reason. They keep going for 50 minutes and then time out. 

Edit: Seems to be caused by an out-of-memory error related to the Radio Astronomy tests.


env-files/torch/skinny.Dockerfile
```diff
@@ -8,7 +8,7 @@
 # --------------------------------------------------------------------------------------
 
 
-ARG BASE_IMG_NAME=python:3.10-slim
+ARG BASE_IMG_NAME=python:3.11-slim
```
matbun, line 11: I'd suggest to downgrade to 3.10 just to be consistent with the other containers. At some point we'll upgrade all of them at the same time

env-files/torch/skinny.Dockerfile
```diff
@@ -35,11 +35,9 @@ RUN uv venv \
     # This is needed by UV to trust all indexes:
     --index-strategy unsafe-best-match \
     # Install packages
-    .[torch] \
+    .[torch,dev] \
     "prov4ml[nvidia]@git+https://github.com/matbun/ProvML@new-main" \
-    pytest \
-    pytest-xdist \
-    psutil
+    lxml rdflib
```
matbun, line 40: this I guess can be reverted back

matbun (changes requested):

lineick (approved):

matbun (approved):

Issue #352, opened 2025-03-19
Cryptic error in config parser when the pipeline key is wrong


matbun:
**Describe the bug**
Running the pipeline with the wrong `pipe_key` causes cryptic error message.

**To Reproduce**

Assuming pipeline 

```yaml
# Pipeline of pipelines 
pop:
  _target_: itwinai.pipeline.Pipeline
  name: parent pipeline
  steps:
    pipeline1:
      _target_: itwinai.pipeline.Pipeline
      name: first pipeline
      steps:
        datagetter:
          _target_: itwinai.tests.dummy_components.FakeGetter
          data_uri: path/to/data
    pipeline2:
      _target_: itwinai.pipeline.Pipeline
      name: second pipeline
      steps:
        datagetter:
          _target_: itwinai.tests.dummy_components.FakeGetter
          data_uri: path/to/data
```

```bash
itwinai exec-pipeline +pipe_key pippo
```

Moreover, it is a common mistake to confuse `+pipe-key` with `+pipe_key`, which at the moment fails silently, resulting in the option to be silently ignored.

```bash
itwinai exec-pipeline +pipe-key pippo
```

**Expected behavior**

Should raise more explicit errors to explain to the user what's wrong.

Anjali-Khantaal:
Getting an error when downloading dataset for MNIST use case (*included the command that worked towards the end*).

Given command in the docs:
`itwinai exec-pipeline +pipe_key=training_pipeline +pipe_steps=[0]`

Error encountered:
```
Error executing job with overrides: ['+pipe_key=training_pipeline', '+pipe_steps=[0]']
Traceback (most recent call last):
  File "/p/project1/intertwin/khantaal1/itwinai/src/itwinai/cli.py", line 721, in exec_pipeline_with_compose
    raise e
  File "/p/project1/intertwin/khantaal1/itwinai/src/itwinai/cli.py", line 713, in exec_pipeline_with_compose
    cfg.steps = [cfg.steps[step] for step in pipe_steps]
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/p/project1/intertwin/khantaal1/itwinai/src/itwinai/cli.py", line 713, in <listcomp>
    cfg.steps = [cfg.steps[step] for step in pipe_steps]
                 ~~~~~~~~~^^^^^^
omegaconf.errors.ConfigKeyError: Key '0' is not in struct
    full_key: training_pipeline.steps.0
[...]

PR #420, opened 2025-07-03
Update exec-pipeline and fix tests

jarlsondre:
Sometimes, `exec-pipeline` gives some rather cryptic errors if you use it incorrectly. This PR changes how these errors are handled by using `raise typer.Exit(1)` for error handling as well as adding more checks and thus more comprehensive error message. 

Additionally, the tests written for the config parser are completely wrong on several accounts. First of all, they use `with raise SystemExit`, which apparently captures the program crashing, meaning that the tests don't do anything. Second, they use for loops instead of `pytest.parametrize`, meaning that you have one huge test instead of separate tests. Thirdly, the config that is given to the `exec-pipeline` command doesn't work, so the tests never actually tested anything since they never got far enough to actually do any of the functionality. There were considerably more mistakes than these, but I will omit them for the sake of brevity... This is also fixed in this PR. 

Related issues: #352

### DIFF
--- a/src/itwinai/utils.py
+++ b/src/itwinai/utils.py
@@ -18,11 +18,14 @@ import sys
 import time
 import warnings
 from collections.abc import MutableMapping
+from omegaconf import DictConfig
 from pathlib import Path
 from typing import Any, Callable, Dict, Hashable, List, Tuple, Type
 from urllib.parse import urlparse
 
+from omegaconf import DictConfig, errors, ListConfig
 import yaml
+import typer
 
 from .loggers import Logger
 
@@ -355,5 +358,45 @@ def time_and_log(
         step=step,
     )
 
-
     return result
+
+
+def filter_pipeline_steps(pipeline_cfg: DictConfig, pipe_steps: List[Any]) -> None:
+    """Filters the steps in the pipeline configuration, `pipeline_cfg`, using `pipe_steps`,
+    and validates the provided `pipe_steps`. Changes the `pipeline_cfg` object in-place. In
+    the event of a validation error, the program is exited, as this is expected to be used
+    from the CLI.
+    """
+    if "steps" not in pipeline_cfg:
+        py_logger.error(
+            "'pipe_steps' were passed but the given pipe key does not contain 'steps'."
+        )
+        raise typer.Exit(1)
+
+    any_numeric_steps = any(isinstance(step, int) for step in pipe_steps)
+    any_string_steps = any(isinstance(step, str) for step in pipe_steps)
+    if any_numeric_steps and isinstance(pipeline_cfg.steps, DictConfig):
+        py_logger.error(
+            "We don't support numbered indices for named steps. Filter steps using their"
+            " names instead of indices."
+        )
+        raise typer.Exit(1)
+    if any_string_steps and isinstance(pipeline_cfg.steps, ListConfig):
+        py_logger.error(
+            "The steps in the given pipeline are unnamed, but some of the `pipe_steps` are"
+            " named. This is not supported. Use indices to filter steps when the steps are"
+            "unnamed."
+        )
+        raise typer.Exit(1)
+
+    try:
+        pipeline_cfg.steps = [pipeline_cfg.steps[step] for step in pipe_steps]
+        py_logger.info(f"Successfully selected steps {pipe_steps}")
+    except errors.ConfigKeyError:
+        py_logger.error(
+            "Could not find all selected steps. Please ensure that all steps exist and that"
+            f"you provided to the dotpath to them. \n\tSteps provided: {pipe_steps}."
+            f"\n\tValid steps: {pipeline_cfg.steps.keys()}"
+        )
+        raise typer.Exit(1)
+
--- a/src/itwinai/utils.py
+++ b/src/itwinai/utils.py
@@ -18,14 +18,13 @@ import sys
 import time
 import warnings
 from collections.abc import MutableMapping
-from omegaconf import DictConfig
 from pathlib import Path
 from typing import Any, Callable, Dict, Hashable, List, Tuple, Type
 from urllib.parse import urlparse
 
-from omegaconf import DictConfig, errors, ListConfig
-import yaml
 import typer
+import yaml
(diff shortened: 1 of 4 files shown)


src/itwinai/utils.py
```diff
@@ -355,5 +357,45 @@ def time_and_log(
...
+
+def filter_pipeline_steps(pipeline_cfg: DictConfig, pipe_steps: List[Any]) -> None:
+    """Filters the steps in the pipeline configuration, `pipeline_cfg`, using `pipe_steps`,
+    and validates the provided `pipe_steps`. Changes the `pipeline_cfg` object in-place. In
+    the event of a validation error, the program is exited, as this is expected to be used
+    from the CLI.
+    """
+    if "steps" not in pipeline_cfg:
+        py_logger.error(
+            "'pipe_steps' were passed but the given pipe key does not contain 'steps'."
```
lineick, line 371: is it intended to use \` for the docstrings but use \' for the errors? (doc strings are prob parsed to markdown by sphinx?)

tests/components/test_config_parsing.py
```diff
@@ -41,121 +40,170 @@ def temp_yaml_files(tmp_path):
...
+    ):
+        pipe_key="my-list-pipeline"
+        path = temp_yaml_files[pipe_key]
+        args = [
+            "exec_pipeline",
+            f"--config-dir={path.parent}",
+            f"--config-path={path.parent}",
+            f"--config-name={path.stem}",
+            f"+pipe_key={pipe_key}",
+            "+pipe_steps=[1]",
```
lineick, line 110: is the indexing still supported? thought this was outdated

lineick (commented):
I dont have good knowledge of pytests but lgtm otherwise (the tests dont seem to do a whole lot though, but thats its own pr i assume)

src/itwinai/utils.py
```diff
@@ -355,5 +357,45 @@ def time_and_log(
...
+        )
+        raise typer.Exit(1)
+
+    any_numeric_steps = any(isinstance(step, int) for step in pipe_steps)
```
matbun, line 375: it would be nice to support slicing, like `+pipe_steps=[1:]` or maybe `+pipe_steps=1:`? Maybe too complicated?

tests/components/test_config_parsing.py
matbun: nono it is supported

tests/components/test_config_parsing.py
```diff
@@ -41,121 +40,170 @@ def temp_yaml_files(tmp_path):
...
+            "itwinai.tests.dummy_components.FakeTrainer.__init__", return_value=None
+        ) as mock_trainer_init,
+        patch(
+            "itwinai.tests.dummy_components.FakeTrainer.execute", return_value=None
+        ) as mock_trainer_exec,
+    ):
+        pipe_key="my-list-pipeline"
+        path = temp_yaml_files[pipe_key]
+        args = [
+            "exec_pipeline",
```
matbun, line 105: is this intended to be exec_pipeline? or should it be itwinai exec-pipeline? or it doesn't matter?

tests/components/test_config_parsing.py
```diff
@@ -41,121 +40,170 @@ def temp_yaml_files(tmp_path):
...
+        patch(
+            "itwinai.tests.dummy_components.FakeTrainer.__init__", return_value=None
+        ) as mock_trainer_init,
+        patch(
+            "itwinai.tests.dummy_components.FakeTrainer.execute", return_value=None
+        ) as mock_trainer_exec,
+    ):
+        path = temp_yaml_files[pipe_key]
+        args = [
+            "exec_pipeline",
```
matbun, line 71: is this intended to be exec_pipeline? or should it be itwinai exec-pipeline? or it doesn't matter?

tests/components/test_config_parsing.py
```diff
@@ -41,121 +40,170 @@ def temp_yaml_files(tmp_path):
...
+            "itwinai.tests.dummy_components.FakeTrainer.__init__", return_value=None
+        ) as mock_trainer_init,
+        patch(
+            "itwinai.tests.dummy_components.FakeTrainer.execute", return_value=None
+        ) as mock_trainer_exec,
+    ):
+        pipe_key="my-dict-pipeline"
+        path = temp_yaml_files[pipe_key]
+        args = [
+            "exec_pipeline",
```
matbun, line 140: is this intended to be exec_pipeline? or should it be itwinai exec-pipeline? or it doesn't matter?

tests/components/test_config_parsing.py
```diff
@@ -41,121 +40,170 @@ def temp_yaml_files(tmp_path):
...
-            mock_trainer_init.assert_called_once_with(
-                lr=0.005, batch_size=64, name="my-trainer"
-            )
+        ) as mock_trainer_init,
+    ):
+        pipe_key = 'my-interpolation-pipeline'
+        path = temp_yaml_files[pipe_key]
+
+        args = [
+            "exec_pipeline",
```
matbun, line 171: is this intended to be exec_pipeline? or should it be itwinai exec-pipeline? or it doesn't matter?

tests/components/test_config_parsing.py
```diff
@@ -41,121 +40,170 @@ def temp_yaml_files(tmp_path):
...
-            monkeypatch.setattr(sys, "argv", args)
-            exec_pipeline()
-
-            mock_preproc_init.assert_called_once_with(max_items=33, name="new-preproc")
+        ) as mock_preproc_init,
+    ):
+        pipe_key = 'my-dict-pipeline'
+        path = temp_yaml_files[pipe_key]
+        args = [
+            "exec_pipeline",
```
matbun, line 196: is this intended to be exec_pipeline? or should it be itwinai exec-pipeline? or it doesn't matter?

tests/components/test_config_parsing.py
jarlsondre: The main change I am making in this PR is to add some error messages, depending on what type you use.

matbun (approved):

Issue #423, opened 2025-07-03
Allow remote pipeline configurations


jarlsondre:
As we would like to create a tutorial where the user doesn't have to clone any libraries, it would be nice to accept a URL for the pipeline configuration in addition to a file. Then the program can check whether the given string is a URL or a path and can act accordingly. 

The idea is that this would allow for something like:

```bash
pip install itwinai
pip install itwinai-mnist-plugin
itwinai exec-pipeline --config-name <github link to pipeline config e.g.>
```

One thing to consider is of course whether it makes sense to use `--config-name` for this or if we should do something like `--config-url` instead.

PR #425, opened 2025-07-04
Allow reading pipeline configurations from online

jarlsondre:
Let's the user pass a `--config-name` that is a URL instead of a filename. The program automatically checks whether it is a URL and in which case, tries to fetch the configuration file from that URL. Expects the file to be a raw file. 

Related Issue: #423


src/itwinai/cli.py
```diff
@@ -660,18 +662,27 @@ def exec_pipeline(
...
-    """Hydra entry function. The hydra.main decorator parses a configuration file
-    (under config_path), which contains a pipeline definition, and passes it to this function
-    as an omegaconf.DictConfig object (called cfg). This function then instantiates and
-    executes the resulting pipeline object.
-    Filters steps if `pipe_steps` is provided, otherwise executes the entire pipeline.
-    For more information on hydra.main, please see
-    https://hydra.cc/docs/tutorials/basic/your_first_app/simple_cli/."""
+    """Hydra entry function.
+
+    The hydra.main decorator parses a configuration file (under config_path), which contains a
```
lineick, line 677: or config_dir?

jarlsondre: Honestly I am not sure, all I did was add some spacing here. You are probably right though

lineick: but config dir is still supported as it is hydra built-in right? so we should mention both?
If not we should hide it from the `itwinai exec-pipeline --help`

PR #435, opened 2025-07-14
add validation_every variable to torch trainer

jarlsondre:
We used to have a variable called `validation_every`, but this has since been removed as it was unused. We still have something called `test_every`, however.

Recently, however, we received a request from a use case to reimplement this behavior, so this PR adds it back to the project.

### DIFF
--- a/src/itwinai/torch/trainer.py
+++ b/src/itwinai/torch/trainer.py
@@ -155,6 +155,10 @@ class TorchTrainer(Trainer, LogMixin):
     validation_dataloader: DataLoader | None = None
     #: PyTorch ``DataLoader`` for test dataset.
     test_dataloader: DataLoader | None = None
+    #: How often to perform the validation step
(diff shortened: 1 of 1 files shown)


src/itwinai/torch/trainer.py
```diff
@@ -1193,50 +1199,52 @@ def train(self) -> None:
...
+                            self.ray_report,
+                            metrics={metric_name: val_metric.item()},
+                            checkpoint_dir=best_ckpt_path or periodic_ckpt_path,
+                        ),
+                        logger=self.logger,
+                        identifier="ray_report_time_s_per_epoch",
+                        step=self.current_epoch,
+                    )
+                else:
+                    self.ray_report(
```
matbun, line 1244: Ray HAS to report at each epoch

matbun (changes requested):
I would not change the suffix to `*_interval`

PR #436, opened 2025-07-14
Containers refactor

### LINKED ISSUE #349: Update Docker container for MNIST use case

### LINKED ISSUE #393: Different version of GCCcore in readthedocs for Vega
I notice that in the docs, we are loading the following modules on Vega:

```
ml --force purge
ml Python/3.11.5-GCCcore-13.2.0 CMake/3.24.3-GCCcore-11.3.0 mpi4py OpenMPI CUDA/12.3
ml GCCcore/11.3.0 NCCL cuDNN/8.9.7.29-CUDA-12.3.0 UCX-CUDA/1.15.0-GCCcore-13.2.0-CUDA-12.3.0
```

However, the version of GCCcore does not match in all the specified modules—some of them use 11.3.0 and some use 13.2.0. Is this something that we should change? I haven't seen any direct problems caused by this, but it seems like something that could cause a bug at some point. It is written on this page: https://itwinai.readthedocs.io/latest/installation/user_installation.html

### LINKED ISSUE #402: Move logic for calculting total number of nodes into the strategy object
Currently, we need the total number of nodes to properly log the epoch times using the `EpochTimeTracker`, but this is calculated using an environment variable and the logic is reused everywhere. This is logic that applies to the distribution of our training, however, which means that it makes more sense to put this in one of those files, e.g. for the strategies. You could for example first check of there is a `SLURM_NNODES` variable set, and if not then you could check the strategy if you could just divide the global world size by the local world size. Maybe also print a warning saying that this could be wrong for heterogenous systems and even check if the local size divides the global one etc.

### LINKED ISSUE #432: Bump py version in containers and migrate to UV where needed
I think both jlab and skinny containers are already using UV. Only slim is not. Another thing I should do is bumping py version to 3.12

matbun:
# Summary

- global bump of python to 3.12
- solves issue with SLURM_NNODES in ray jobs (Ray overrides SLURM_NNODES  and cannot be used to log epochtime)
  - updates the epochtime tracker accordingly
  - updates plotting for scalability report accordingly
- adds ARM container CI for itwinai
- updates k3s dependency in dagger CI
- updates env modules for JSC to Stages/2025
- updates env modules for Vega (moves to CUDA 12.6)
- new Vega scripts for MNIST
- updates containers
  - for lumi
  - updates python version in containers
  - uses uv to install py deps in slim container
  - fixes some uv install problems in the skinny container (uv should not install in the global namespace)
  - removed some old dockerfiles
- updates dependecies: torch (2.4 -> 2.6), deepspeed (0.15 -> 0.16.8)
- adds some checks to test the distributed setup (e.g., network), useful to test HPC setup with containers with minimal scripts to help during debugging

...and other things I don't remember

tested on LUMI, JSC, and Vega

---

**Related issue :**

- Fixes #432
- Fixes #402 -- the problem was solved by avoiding to rely on the number of nodes (global world size is a more robust approach)
- Fixes #393
- Fixes #349


ci/src/main/main.py
```diff
@@ -115,16 +118,32 @@ async def build_container(
...
     ) -> Self:
         """Build itwinai container image from existing Dockerfile"""
         if build_args:
             build_args = [
                 BuildArg(name=arg_couple.split("=")[0], value=arg_couple.split("=")[1])
                 for arg_couple in build_args
             ]
+        if build_arm:
+            # Build also for ARM
+            print("INFO: Building container for linux/arm64 platform")
```
jarlsondre, line 134: Does it not make sense to use a logger for the CI stuff?

src/itwinai/cli.py
```diff
@@ -405,6 +405,39 @@ def sanity_check(
...
+    """
+    from itwinai.tests.distributed import test_cuda, test_gloo, test_nccl, test_ray, test_rocm
+
+    match platform:
+        case "nvidia":
+            test_cuda()
+        case "amd":
+            test_rocm()
+        case _:
+            type.echo("Unrecognized platform!")
```
jarlsondre, line 429: `typer`?

src/itwinai/cli.py
```diff
@@ -405,6 +405,39 @@ def sanity_check(
...
+
+    match launcher:
+        case "torchrun":
+            test_gloo()
+            test_nccl()
+        case "ray":
+            test_ray()
+        case _:
+            type.echo("Unrecognized launcher!")
```
jarlsondre, line 438: `typer`?

.vscode/extensions.json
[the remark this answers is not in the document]
matbun: When I find a useful ext I add it so other ppl may benefit from it

ci/src/main/main.py
matbun: Mmm no, here the print is there just to enrich a bit more the logs on Dagger UI

src/itwinai/cli.py
matbun: right

jarlsondre (approved):

env-files/torch/jupyter/Dockerfile
```diff
@@ -11,6 +11,16 @@
 # https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#jupyter-pytorch-notebook
 # https://github.com/jupyter/docker-stacks/wiki
 
+
+
+
+
+# WARNING: THIS DOCKERFILE IS OLD AND NOT MAINTAINED ANYMORE! LEFT HERE FOR REFERENCE
```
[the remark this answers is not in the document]
matbun: I think you're right. I can cleanup a bit

use-cases/mnist/torch/config.yaml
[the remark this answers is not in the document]
matbun: `experiment_name` is there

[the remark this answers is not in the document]
lineick: but it says line 43 experiment_name was removed.
The problem with just having the num_workers being defined outside is just a bit weird, would add the trials and worker resources then too.

jarlsondre (dismissed):

PR #439, opened 2025-07-16
MLflow logging and GPU Profiling Rework

lineick:
MLflow logging has been fixed and improved, to allow logging on worker level with worker runs nesting into the parent train run. For ray HPO the nesting goes one level deeper with a tune run, nesting train runs, nesting workers. 

GPU Profiling now integrates with the itwinai loggers (Adding gpu stats on worker level with side processes)
Updated GPU plotting, for new mlflow-based gpu profiling per worker.


src/itwinai/torch/trainer.py
matbun, line 1: This is a comment on #431 . Please provide a default value to `experiment_name` argument when not provided by the user

lineick: I kept that in mind and checked it in the code it is already creating an exp-name if none is given, i will check the code for errors but it is there

lineick: ok so the problem was not properly handled, it now uses the same default experiment_name used for the loggers, and if the given experiment_names differ it will raise an error.

So in the case of nothing given everything will work, as it will use the "unnamed-experiment" name.

src/itwinai/cli.py
```diff
@@ -33,7 +33,8 @@
 from omegaconf import DictConfig
 from typing_extensions import Annotated
 
-from itwinai.utils import COMPUTATION_DATA_DIR, EPOCH_TIME_DIR, GPU_ENERGY_DIR
+from itwinai.loggers import BASE_EXP_NAME
```
matbun, line 36: this innocent looking line is actually introducing overhead every time we run `itwinai` in the terminal. It would be better to move as many imports as possible into the dedicated command functions, especially in the case of itwinai.loggers which imports all the logger frameworks in background. I'd suggest to either create a dedicated lightweight itwinai module like `itwinai.constants`, or using None as default, and handle it inside the function

src/itwinai/cli.py
```diff
@@ -33,7 +33,8 @@
...
+from itwinai.utils import COMPUTATION_DATA_DIR, EPOCH_TIME_DIR
```
matbun, line 37: `itwinai.utils` is a bit more lightweight, but this line could benefit for the same refactor suggested above

src/itwinai/torch/monitoring/monitoring.py
```diff
@@ -88,6 +90,30 @@ def profile_gpu_utilization(
...
+        identifier="num_global_gpus",
+        kind="param",
+        force=True,
+    )
+    logger.log(
+        item=probing_interval,
+        identifier="probing_interval",
+        kind="param",
+        force=True,
+    )
```
matbun, line 116: I don't get the reason for the `force` argument

matbun (changes requested):
I'd feel comfortable approving this PR once the conversation on #431 is clarified and, if needed, the changes are implemented here

src/itwinai/cli.py
```diff
@@ -157,6 +157,10 @@ def generate_py_spy_report(
 
 @app.command()
 def generate_scalability_report(
+    experiment_name: Annotated[
+        str,
+        typer.Option(help="The name of the experiment to use for the GPU data report."),
```
jarlsondre, line 162: Is this name only for the GPU data report and not the others? In which case, maybe it should be called something like `gpu_data_experiment_name` or `mlflow_experiment_name`? Or something else, but just something that indicates that it only belongs to one of them and not all of them?

jarlsondre (approved):
Simply outstanding

use-cases/mnist/torch/runall.jsc.sh
matbun, line 1: uncomment ds and hvd

use-cases/mnist/torch/slurm.jsc.sh
```diff
@@ -183,9 +182,9 @@ export ITWINAI_LOG_LEVEL="DEBUG"
 # Launch training
 if [ "$DIST_MODE" == "ddp" ] ; then
   echo "DDP training: $TRAINING_CMD"
-  torchrun-launcher "$TRAINING_CMD"
- 
-  separation
+  # torchrun-launcher "$TRAINING_CMD"
+
+  # separation
```
matbun, line 187: uncomment

use-cases/mnist/torch/runall.jsc.sh
lineick: left it outcommented

jarlsondre (dismissed):
you've attained divinity.

src/itwinai/torch/distributed.py
```diff
@@ -1025,6 +1066,22 @@ def gather(self, tensor: torch.Tensor, dst_rank: int = 0) -> Optional[List[torch
...
+        Args:
+            obj (Any): object to broadcast to all workers.
+            src_rank (int): the rank that broadcasted
+
+        Returns:
+            Any: broadcasted object.
+        """
+        # https://horovod.readthedocs.io/en/stable/_modules/horovod/torch/functions.html#broadcast_object
+        return self.hvd.broadcast_object(obj, root_rank=src_rank)
+
```
matbun, line 1082: if this is failing when obj is None, maybe add a warning to help the user debugging the error?

jarlsondre (approved):
Pristine

PR #447, opened 2025-08-04
Full integration of scalability analysis with mlflow

lineick:
This PR removes the EpochTimeTracker and tracks epoch time instead via `mlflow`.
The call times from torch profiling are also stored to `mlflow` now.
The scalability report no longer relies on a `scalabilty-metrics` directory, instead all information is tracked via `mlflow`, allowing for more resistance against crashes etc. and unifying the metrics at one place only.

This integration also allows for the generation of scalability plots from remote `tracking-uri`.

Additional:
- fixed sorting of workers for gpu plots
- fixed error if no log_on_workers is given
- fixed racing condition for experiment creation

### DIFF
--- a/src/itwinai/scalability_report/data.py
+++ b/src/itwinai/scalability_report/data.py
@@ -104,7 +104,7 @@ def read_epoch_time_from_mlflow(
     Args:
         experiment_name (str): Name of the MLflow experiment to read from.
         run_names (List[str] | None): Name of the runs to read metrics from. If empty, all runs
-        in the experiment will be considered.
+            in the experiment will be considered.
     Returns:
         pd.DataFrame | None: A DataFrame containing the concatenated data from all epoch time
         metrics in the given runs of the experiment.
--- a/src/itwinai/torch/mlflow.py
+++ b/src/itwinai/torch/mlflow.py
@@ -347,13 +347,13 @@ def get_runs_by_name(
         # get all run IDs from the experiment that are not-nested
         runs = mlflow_client.search_runs([experiment_id])
         runs = [run for run in runs if "mlflow.parentRunId" not in run.data.tags]
+        return runs
 
-    else:
-        runs = []
-        # get all runs in the experiment
-        for run_name in run_names:
-            runs += mlflow_client.search_runs(
-                experiment_ids=[experiment_id],
-                filter_string=f"run_name='{run_name}'",
-            )
+    runs = []
+    # get all runs in the experiment
+    for run_name in run_names:
+        runs += mlflow_client.search_runs(
+            experiment_ids=[experiment_id],
+            filter_string=f"run_name='{run_name}'",
+        )
     return runs


lineick:
> One thing to test, if not already done, is whether plots can be generated using a remote MLflow server, where the tracking URI is not a local abs path, but a URL

Yes, will do, have to get it to run first, there are quite a lot of dependencies and more changes than one might expect for such simple functionality.

src/itwinai/scalability_report/data.py
```diff
@@ -14,60 +14,168 @@
...
+
+    return pd.concat(epoch_time_dataframes)
+
+
+def read_gpu_metrics_from_mlflow(
+    mlflow_client: MlflowClient,
+    experiment_name: str,
+    run_names: List[str] | None = None,
+) -> pd.DataFrame | None:
+    """Reads and validates GPU metrics from a mlflow experiment and combines them into a
```
jarlsondre, line 148: "an mlflow ..."

use-cases/mnist/torch/runall.vega.sh
matbun, line 1: would be nice to restore the previous state of this file

use-cases/mnist/torch/slurm.vega.sh
```diff
@@ -225,10 +224,10 @@ if [ "${DIST_MODE}" == "ddp" ] ; then
...
   torchrun-launcher "${TRAINING_CMD}"
 
-  separation
-
-  decho -e "\nLaunching DDP strategy with Ray"
-  ray-launcher "${TRAINING_CMD}"
+  # separation
+  #
+  # decho -e "\nLaunching DDP strategy with Ray"
+  # ray-launcher "${TRAINING_CMD}"
```
matbun, line 230: why is this commented out?

src/itwinai/scalability_report/plot.py
```diff
@@ -118,9 +119,9 @@ def relative_epoch_time_speedup_plot(
 
     Args:
         avg_epoch_time_df (pd.DataFrame): A DataFrame containing:
-            - "workers": Number of workers used in the training process.
+            - "num_global_gpus": Number of GPUs used in the training process.
```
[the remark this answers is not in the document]
lineick: Hm good point. I think a worker can not have more than one gpu. At least with none of our libraries. In our case it should be correct to say num_global_gpus though. I will check

src/itwinai/torch/mlflow.py
```diff
@@ -101,8 +104,127 @@ def teardown_lightning_mlflow() -> None:
...
+    if not first_level_children:
+        cli_logger.warning(
+            f"No child runs found for run ID {run.info.run_id} in experiment {experiment_id}."
+        )
+        return worker_profiling_averages
+
+    leaf_nodes = []
+    for child in first_level_children:
+        second_level_children = _children(child.info.run_id)
+        leaf_nodes += second_level_children if second_level_children else [child]
```
jarlsondre, line 197: ahh very smart

src/itwinai/torch/trainer.py
```diff
@@ -331,11 +329,13 @@ def _detect_distributed_strategy(self, strategy: str) -> TorchDistributedStrateg
...
-        enough_resources = distributed_resources_available() or ray_cluster_is_running()
+        dist_resources = distributed_resources_available()
+        ray_cluster_running = ray_cluster_is_running()
+        enough_resources = dist_resources or ray_cluster_running
         py_logger.debug(
             f"Enough resources? {enough_resources} "
-            f"(distributed_resources_available: {distributed_resources_available()}) "
-            f"(ray_cluster_is_running: {ray_cluster_is_running()})"
+            f"(distributed_resources_available: {dist_resources}) "
+            f"(ray_cluster_is_running: {ray_cluster_running})"
```
jarlsondre, line 338: Why the parentheses around these? Not very important or anything but could probably just use a comma or something?

src/itwinai/scalability_report/data.py
```diff
@@ -10,120 +10,191 @@
...
+    experiment_name: str,
+    run_names: List[str] | None = None,
+) -> pd.DataFrame | None:
+    """Reads and validates epoch time metrics from a mlflow experiment and combines them into a
+    single DataFrame.
+
+    Args:
+        experiment_name (str): Name of the MLflow experiment to read from.
+        run_names (List[str] | None): Name of the runs to read metrics from. If empty, all runs
+        in the experiment will be considered.
```
jarlsondre, line 107: indentation for the second line...

src/itwinai/torch/mlflow.py
```diff
@@ -101,8 +104,124 @@ def teardown_lightning_mlflow() -> None:
...
+
+    for child in first_level_children:
+        second_level_children = _children(child.info.run_id)
+        # exists for ray runs
+        if second_level_children:
+            for grand_child in second_level_children:
+                if "epoch_time_s" in grand_child.data.metrics:
+                    epoch_time_runs.append(grand_child)
+        elif "epoch_time_s" in child.data.metrics:
+            epoch_time_runs.append(child)
```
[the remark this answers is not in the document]
lineick: no

matbun (approved):

PR #450, opened 2025-08-11
Initial set-up for the runner

okrochak:
Ideally, the status of the runner should be monitored with itwinai commands. 

The output of SLURM job is parsed via the runner and can be read on GitHub actions log directly, which is in accordance to open-source principles in contrast to currently employed Dagger container runner which requires explicit access to read the error logs.

### DIFF
--- a/.github/workflows/juwels-runner.yml
+++ b/.github/workflows/juwels-runner.yml
@@ -1,0 +1,58 @@
+name: Test Juwels Runner
+
+on:
+  push:
+    branches:
+      - 'main'
+
+jobs:
+  run-command:
+    runs-on: [self-hosted, linux]  # adjust labels if you used custom ones
+    env:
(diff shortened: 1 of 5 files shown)


okrochak:
> It is not true that with Dagger you cannot see the logs. It is actually very well designed and they can be accessed from here: https://dagger.cloud/itwinai
> 
> If needed, I can add you using your github email

Where did I say that you cannot read the logs? I've said:
"requires explicit access to read the error logs", and this is true. 

With my proposed implementation, all logs will be visible directly on GitHub for anyone to see. And I won't need to ask anyone to make an account and give them permission

matbun:
> @matbun this should be ready to merge now. I think the cleanest way to have this alternative workflow "juwels-runner" which will only be triggered for selected branches

Considering that this workflow would be triggered only on `juwels-runner` branch, is there a reason to have it on main (or on other branches) as well?

r-sarma (commented):
Looks good to me!

PR #456, opened 2025-08-20
Create itwinai run command

jarlsondre:
This PR adds an `itwinai run` command that can be used to launch an entire workflow, with installation, just from a single config file. All you need to do is specify a configuration file with your pipeline, some plugins that you need to install and a `slurm_config`. An example:

```bash
itwinai run -c config.yaml
```

### DIFF
--- a/use-cases/mnist/torch/config.yaml
+++ b/use-cases/mnist/torch/config.yaml
@@ -37,6 +37,17 @@ ray_gpus_per_worker: 1  # can be 0.0 < 1.0 as long as num_workers set to 1
 ray_cpus_per_worker: 3
 ray_num_trials: 1 # how many HPO trials
 
+plugins:
+- git+https://github.com/matbun/itwinai-mnist-plugin.git
+
+slurm_config:
+  job_name: mnist-job
+  memory: 64G
+  mode: single
+  num_nodes: 8
+
+  # pre_exec_file: ../../../src/itwinai/slurm/system-base-scripts/juwels_pre_exec.sh
+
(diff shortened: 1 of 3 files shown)


docs/examples/complete-workflow-mnist-config.yaml
matbun, line 1: For the sake of maintainability, I'd probably move this under `use-cases/mnist/run-example` (or similar name) to keep mnist-related stuff close to each other.

docs/getting-started/complete-workflow-example.rst
matbun, line 1: this is good, but it would be nice to reflect this also in the [Quick Start code block](https://itwinai.readthedocs.io/latest/#quick-start) on the readthedocs home, showing how to run a whole workflow in one command

use-cases/mnist/torch/config.yaml
```diff
@@ -37,6 +37,17 @@ ray_gpus_per_worker: 1  # can be 0.0 < 1.0 as long as num_workers set to 1
...
+# plugins:
+# - git+https://github.com/matbun/itwinai-mnist-plugin.git
+
+slurm_config:
+  job_name: mnist-job
+  memory: 64G
+  mode: single
+  num_nodes: 8
+
+  pre_exec_file: src/itwinai/slurm/system-base-scripts/juwels_pre_exec.sh
```
matbun, line 49: It's be nice to have a URL here, to make this more portable and path-agnostic

matbun (changes requested):

PR #458, opened 2025-09-04
Reducing WC for JOSS

matbun:
# Summary

- Summarised `Summary` and `Statement of Need` sections, trying to keep only what is strictly necessary.
- Converted some footnotes to direct links to save few words

---

**Related issue :**


r-sarma (approved):
Looks good to me!

PR #476, opened 2025-12-12
Add a generic section in the installation document

### LINKED ISSUE #468: add a generic section in the installation document
In these two pages:
https://itwinai.readthedocs.io/latest/installation/user_installation.html
https://itwinai.readthedocs.io/latest/installation/developer_installation.html

1. would you consider to add the requirements (or required modules) for generic HPC? So we know which packages/versions are necessary and must be load from ml; which is optional and might be able to load later via pip.

In our university, we use [Rivanna](https://www.rc.virginia.edu/userinfo/hpc/). We have a very different module list than these provided clusters. A list of available modules can be found [here](https://www.rc.virginia.edu/userinfo/hpc/software/complete-list/).

For Juelich (JSC), PyTorch:
```shell
ml --force purge
ml Stages/2025 # what is this for? Rivanna don't have this package.
ml GCC OpenMPI git # any specific requirements to package version? in Rivanna, openmpi is included in gompi.
ml CUDA/12 # I guess the cuda version depends on PyTorch/Tensorflow, but it would be good to state if it has to be 12, if there's sub-version requirement, or if it works for cuda 11 as well.
ml cuDNN
ml MPI-settings/CUDA # what's the difference of this cuda to the cuda above?
ml Python CMake HDF5 PnetCDF libaio mpi4py
ml libaio # did not find this in rivanna
```

For Juelich (JSC), TensorFlow:
I saw Stages have different version (2024), and many dependencies have detailed version requirement.

Vega:
This looks close to the packages provided by Rivanna, but are these version strict? I also noticed additonal dependencies of nccl.

LUMI:
Could you provide some brief explaination why Python virtual environments are discouraged in favour of containers on this cluster?

My command for installation for your reference:
```shell
ml --force purge
ml gcc gompi cudnn python cmake hdf5 pnetcdf mpi4py git uv nccl cuda
```

2. Since dependencies are the same for PyTorch and TensorFlow in Vega, I would suggest to leave one tab.

ref https://github.com/openjournals/joss-reviews/issues/9409

matbun:
# Summary

TODO:

- [x] Add generic SW dependencies list with version ranges (e.g., CUDA>=12.4)
- [x] Ping DeepSpeed and Horovod versions
- [x] Fix typo in itwinai user installation page
- [x] [I would recommend to add a section in the document advising something like](https://github.com/interTwin-eu/itwinai/issues/468#issuecomment-3647709126) "while itwinai do not have requirements for certain version of CUDA, PyTorch, TensorFlow or other packages, it is common to face problems while deploying on a new system. We would encourage the user do some trial with difference versions of these software. We will (or will not, or depends on availability, will) support for these systems. Users can create a new issue or pull request to report their experience on their own systems."

Docs can be visualized here: https://itwinai--476.org.readthedocs.build/476/installation/user_installation.html

---

**Related issue :**

Fixes #468

### DIFF
--- a/docs/installation/hpc_modules.rst
+++ b/docs/installation/hpc_modules.rst
@@ -6,7 +6,7 @@ documentation for specific guidance.
 
 Find below the modules you should load on the supercomputers where we tested itwinai, depending
 on whether you want ``PyTorch`` or ``TensorFlow`` support. If you are deploying itwinai on a
-different HPC system, please refer to the *Other HPCs* tab.
+different HPC system, please refer to the **Other HPCs** tab.
 
 
 .. tab-set::


docs/installation/hpc_modules.rst
r-sarma: Could you add me as co-author to this file? Thanks!

r-sarma (commented):
Rest look good to me! But I wait for the review and feedback from @xiazeyu

PR #502, opened 2026-07-21
Update to Model Hub integration

r-sarma:
This PR is an update to the RI-SCALE Model Hub integration with itwinai, which was a CLI-based implementation.
1. Now the AI Model Hub is added as a feature, allowing the abstraction of the backend, when pulling models. Users are able to define this already in the configuration file.
2. The model pulling API is also added, which allows the users to pull models from the Model Hub and launch inference.


docs/how-it-works/model-hub/explain_model_hub.rst
okrochak, line 1: GitHub preview shows me a lot of text being underlined, which I think is not so good for readability.

src/itwinai/torch/model_hub/utils.py
okrochak, line 1: Maybe it's better to put this function into `features.py` ? I feel like 1 file with 1 function is too little, but it's up to you

src/itwinai/torch/inference.py
okrochak, line 1: Again, a subjective thing, but maybe it's better to put `ModelHubModelLoader` into `model-hub` folder and import it in `inference.py`, to more clearly manage Model Hub code.

tutorials/model-hub/torch-tutorial-model-hub/README.md
okrochak, line 1: I've tried to run this tutorial with a fresh install of `itwinai` from this branch `modelhub-in-trainer`. I don't think the .env file is uploaded, and when I try to use the .env file I have configured myself for the model hub, I still get an error even though `.env` file was parsed: 

```Loaded environment from .env file in current directory
[ERROR]: API token not provided. Set it via:
  - --api-token option
  - HYPHA_TOKEN environment variable
  - HYPHA_TOKEN in .env file```

I will try to look into this deeper

**Review by okrochak — COMMENTED** [review]
I've left some comments about the modelhub-related code organization. I've tried to run the tutorial, also with my own `.env` file, but didn't manage to (see the last comment).

**Comment by r-sarma** [comment]
@matbun Could you please review this?

**Comment by matbun** [comment]
> @matbun Could you please review this?

I was waiting for you to iterate on Alex's comments, but I can review it in parallel. Will do by the end of the week!

File: tests/conftest.py [hunk 1]
```diff
@@ -12,6 +12,8 @@
 
 import pytest
 
+os.environ.setdefault("MLFLOW_ALLOW_FILE_STORE", "true")
+
```
matbun: This change is making the CI pass, but it is hiding the root cause, which is being solved by #505 
If the user doesn't remember to set this env var the code will break, which is something that the tests are not able to detect anymore.

I would suggest removing this, merging #505 first, and rebasing on main to bring in the fix

src/itwinai/torch/model_hub/feature.py
```diff
@@ -0,0 +1,49 @@
...
+from itwinai.torch.model_hub.backends import get_backend
+from itwinai.torch.model_hub.manifest import write_manifest
+from itwinai.torch.model_hub.utils import has_internet_connection
+
+
+class ModelHubFeature:
+    def __init__(self, config: dict):
+        self.config = config or {}
+        self.enabled = self.config.get("enabled", False)
+        self.final_checkpoint_name = self.config.get("final_checkpoint_name", "best_model")
```
matbun: `final_checkpoint_name` is configurable here, but `trainer.py:1280` hardcodes the directory it
passes in:

```python
best_ckpt_dir = Path(self.checkpoints_location) / "best_model"
```

So if a user sets `final_checkpoint_name: my_ckpt`, `on_training_end` hits the
`ckpt_dir.name != self.final_checkpoint_name` guard on line 29, returns early, and **the whole
feature silently does nothing** — no manifest, no upload, no message. The knob is only ever
correct at its default value.

Two ways out: drop the option and hardcode `"best_model"` in both places, or (better) have the
trainer ask the feature for the name it wants. The latter also removes the guard entirely:

```python
# trainer.py
if self.strategy.is_main_worker and self._model_hub.enabled:
    best_ckpt_dir = Path(self.checkpoints_location) / self._model_hub.final_checkpoint_name
    if best_ckpt_dir.exists():
        self._model_hub.on_training_end(self, best_ckpt_dir)
```

Worth noting `"best_model"` is also hardcoded at `trainer.py:1235` in the `save_checkpoint`
call, so a real fix probably wants a single constant both sites share.

src/itwinai/torch/model_hub/backends/itwinai_hub.py
```diff
@@ -0,0 +1,13 @@
...
+
+from .base import BaseBackend
+
+
+# TODO: Can be extended to add other backends, e.g. HuggingFace
+class AIModelHubBackend(BaseBackend):
+    def upload(self, model_dir: Path):
+        subprocess.run(
+            ["itwinai", "upload-model-to-hub", str(model_dir)],
+            check=False,
```
matbun: `check=False` means a non-zero exit from `itwinai upload-model-to-hub` never raises, right? in that case, this would make 
`_safe_upload`'s `except Exception` at `feature.py:47` unreachable. so on a failed upload the
user never sees *"Model Hub upload failed…"* or the *"You can re-upload later with…"* hint, and
training reports success.

src/itwinai/torch/inference.py
```diff
@@ -21,9 +21,61 @@
...
+            raise ConnectionError(
+                "No internet connection: cannot reach the Model Hub to pull "
+                f"model '{self.model_id}'."
+            )
+
+        file_path = self.file_path or discover_weights_file(self.base_url, self.model_id)
+        dst_dir = Path("tmp") / "modelhub_downloads" / self.model_id
+        ckpt_path = download_file(self.base_url, self.model_id, file_path, dst_dir)
+
+        checkpoint = torch.load(ckpt_path, weights_only=False)
```
matbun: I would suggest to set `weights_only=True` here, as it seems we are loading only the weights of the model. This has some important [security implications](https://docs.pytorch.org/docs/2.13/notes/serialization.html#torch-load-with-weights-only-true) considering we are unpicking a file we downloaded from the internet from a remote source which could have been compromised.

src/itwinai/torch/inference.py
```diff
@@ -21,9 +21,61 @@
...
+        if self.model_class is None:
+            raise ValueError(
+                "model_class is required: Model Hub checkpoints store weights "
+                "as a raw state_dict, not a pickled model object."
+            )
+        model = self.model_class()
+        if isinstance(checkpoint, dict) and "model_state_dict" in checkpoint:
+            model.load_state_dict(checkpoint["model_state_dict"], strict=False)
+        else:
+            model.load_state_dict(checkpoint, strict=False)
```
matbun: With `strict=False`, `load_state_dict` returns silently when no key matches at all. Combined
with pulling weights from a remote hub and requiring the user to supply `model_class` by hand,
this is the exact scenario where a mismatch is likely, and the result is a randomly
initialised model that runs inference and returns plausible-looking garbage... But if this is too strict please ignore this comment

matbun (changes requested):
Nice PR, it definitely improves a lot the integration with the AI model hub!

Important points (the order is random):

- I would suggest moving the `model_hub` package from `itwinai/torch/model_hub` to one level up as `itwinai/model_hub`. The reason is that itwinai is naturally designed to support multiple frameworks, and `model_hub` should be considered at the same level as pytorch. Also considering that model hub does not strictly depend on torch. This also simplifies the import from `itwinai.torch.model_hub` to `itwinai.model_hub`
- As already mentioned inline, I would remove the change in `conftest.py` and rebase on #505. This way we don't risk of having broken code while the CI still passes.
- No test touches `model_hub/*`, `ModelHubModelLoader`, or the CLI changes. I would suggest adding some tests.
- The Model Hub pull runs on every rank. `TorchPredictor.execute` calls `self.model = self.model()` without a rank guard, so all workers download the same checkpoint to the same CWD-relative `tmp/modelhub_downloads/<model_id>/` path and race on the write. On HPC that path is usually the submit directory on a shared filesystem, so it's a cross-node corruption race, not just wasted bandwidth. The guard can't go inside `ModelHubModelLoader` since a `ModelLoader` is a bare callable with no strategy handle, but `execute()` method owns `self.strategy`: download under if `self.strategy.is_main_worker`: then `self.strategy.barrier()`, since every rank still needs the file before `distribute_model()`


Other points:

- As discussed in a [previous thread](https://github.com/interTwin-eu/itwinai/pull/490#discussion_r2762841651) this could be a good opportunity of moving some code from `upload_to_model_hub` and `_load_env_file` to the new `model_hub` package to save some lines in `cli.py` . A similar reasoning is true for other functions we added in `cli.py`... If you agree I will open a separate issue for cleaning this up as well.

matbun:
Hi @r-sarma, I saw your last commits but I haven't started a review yet because I don't know if you are already done. Don't hesitate to re-request a review when ready. No rush, though

r-sarma:
> Hi @r-sarma, I saw your last commits but I haven't started a review yet because I don't know if you are already done. Don't hesitate to re-request a review when ready. No rush, though

Mostly done but I am adding some tests. I will re-request once its ready.

docs/how-it-works/model-hub/explain_model_hub.rst
r-sarma: As we had discussed, it seems to look fine for me. Perhaps its a matter of how you view the file?

src/itwinai/torch/model_hub/utils.py
r-sarma: I have moved it now, as per your suggestion.

src/itwinai/torch/inference.py
r-sarma: This is a `torch` component, hence I had it here. But now everything related to model hub is moved out one level higher. I have moved it for now, as per your suggestion. But we can decide here, if you think otherwise.

tutorials/model-hub/torch-tutorial-model-hub/README.md
r-sarma: The issue was related to the name of the environment variable. It should be `HYPHA_API_TOKEN`, which has been updated now.

tests/conftest.py
r-sarma: Yes, you are right. It has been rebased now.

src/itwinai/torch/model_hub/feature.py
r-sarma: Good catch! As per your suggestion, the checkpoint name is now configurable.

src/itwinai/torch/model_hub/backends/itwinai_hub.py
r-sarma: Agreed, I have changed this option to `True` now.

src/itwinai/torch/inference.py
r-sarma: Good point - I have modified as suggested.

src/itwinai/torch/inference.py
r-sarma: I think its a good suggestion. Changed to `strict=True`.

r-sarma:
> Nice PR, it definitely improves a lot the integration with the AI model hub!
> 
> Important points (the order is random):
> 
> * I would suggest moving the `model_hub` package from `itwinai/torch/model_hub` to one level up as `itwinai/model_hub`. The reason is that itwinai is naturally designed to support multiple frameworks, and `model_hub` should be considered at the same level as pytorch. Also considering that model hub does not strictly depend on torch. This also simplifies the import from `itwinai.torch.model_hub` to `itwinai.model_hub`
> * As already mentioned inline, I would remove the change in `conftest.py` and rebase on #505. This way we don't risk of having broken code while the CI still passes.
> * No test touches `model_hub/*`, `ModelHubModelLoader`, or the CLI changes. I would suggest adding some tests.
> * The Model Hub pull runs on every rank. `TorchPredictor.execute` calls `self.model = self.model()` without a rank guard, so all workers download the same checkpoint to the same CWD-relative `tmp/modelhub_downloads/<model_id>/` path and race on the write. On HPC that path is usually the submit directory on a shared filesystem, so it's a cross-node corruption race, not just wasted bandwidth. The guard can't go inside `ModelHubModelLoader` since a `ModelLoader` is a bare callable with no strategy handle, but `execute()` method owns `self.strategy`: download under if `self.strategy.is_main_worker`: then `self.strategy.barrier()`, since every rank still needs the file before `distribute_model()`
> 
> Other points:
> 
> * As discussed in a [previous thread](https://github.com/interTwin-eu/itwinai/pull/490#discussion_r2762841651) this could be a good opportunity of moving some code from `upload_to_model_hub` and `_load_env_file` to the new `model_hub` package to save some lines in `cli.py` . A similar reasoning is true for other functions we added in `cli.py`... If you agree I will open a separate issue for cleaning this up as well.

Thanks for the suggestions! I found them quite useful to improve this PR.
- The model_hub package has been moved one level higher now.
- The rebasing has been done - so the MLflow changes are now included.
- Tests have been added to test individual components and also the complete integration.
- Good suggestion on the potential issue with Model Hub pull running on multiple ranks. Now `is_main_worker` check has been used on the `execute()` method to only do this on the main worker.
- On the CLI, I have now moved the Model Hub-related functions to the `itwinai/model_hub` directory. Only one wrapper function remains in the main itwinai `cli.py`. I think its a good idea to also do this for the other functions. I could help you there - let me know.

r-sarma:
@matbun @okrochak Thanks guys for reviewing this PR - your comments were very useful. I think its ready now for the next round of review.

docs/tutorials/model-hub/model_hub_tutorial.rst
```diff
@@ -0,0 +1,33 @@
...
+.. include:: ../../../tutorials/model-hub/torch-tutorial-model-hub/README.md
+   :parser: myst_parser.sphinx_
+   :start-line: 4
+
+Run the training pipeline (trains the model and, as per ``model_hub`` in ``config.yaml``,
+pushes the best checkpoint to the Model Hub):
+
+.. code-block:: bash
+
+   itwinai exec-pipeline +pipe-key training_pipeline
```
okrochak, line 12: For me it works with `+pipe-key=training_pipeline`, same fopr l.18. Throws an error without the equality sign.

okrochak: Also I think it's worth mentioning in the tutorial and/or docs explicitly that `modelhub` (and `torch` too I think) are necessary for this to work, so the user needs to make sure the dependencies are there in his .venv or run `uv sync --extra torch --extra modelhub`.

tutorials/model-hub/torch-tutorial-model-hub/README.md
okrochak, line 1: I think this is a very good explanation but it might be confusing because there is also a file in the docs (docs/tutorials/model-hub/model_hub_tutorial.rst) . So either they have to be consistent (i.e. +/- a copy of each other) or one of them references the other

matbun: Yes I agree! I would probably reference the docs from here, trying to keep this file minimal

matbun (dismissed):
Apart from the same [point](https://github.com/interTwin-eu/itwinai/pull/502#discussion_r4027025870) that @okrochak already made, it looks good to me!

r-sarma:
Hi @matbun @okrochak I made some changes to the documentation as per your suggestions. Could you check if this looks ok?

matbun (approved):
LGTM

okrochak:
LGTM, thank you Rakesh. @matbun I am unfortunately on sick leave this week, so I will have to look at the Claude Skill MR later.

okrochak (approved):

Issue #156, opened 2024-06-17
Easily access datasets on Rucio data lake


matbun:
Add a Python function capable of translating a namespaced Rucio dataset/file to the absolute path on the local filesystem of the datacenter (e.g., HPC) on which the code is currently running. 

Sth like `namespace_to_path('jdoe:physics_dataset')` returning:

- `'/dacache/slling.si/.../physics_dataset'` when on HPC1
- `'/other/path/.../physics_dataset'` when on HPC2

The dataset can or cannot be on the HPC:

- When the dataset is available on the local RSE, return a list of paths to the dataset files (or just the path to the root directory of the dataset, if you prefer).
- When the dataset is not there, create a Rucio rule for _async_ copy of the dataset and raise a custom exception to inform the user that the dataset is not present at the moment and that the job cannot continue, although a rule has been created.

How to proceed;

- Create a `rucio.py` module under [`src/itwinai/`](https://github.com/interTwin-eu/itwinai/tree/main/src/itwinai) to store the python function meant to convert a rucio dataset to the absolute path on the local RSE
- Add tests in a `test_rucio.py` file under [`tests/`](https://github.com/interTwin-eu/itwinai/tree/main/tests)

Once this is done, we will integrate it with other itwinai modules (e.g., config parser and CLI)

garciagenrique:
Hello @matbun,

After speaking with few people at CERN, there are two "main" way to interact with RUCIO data. 

1. Download the desired dataset into the localhost.
2. Make a replication rule so that the files are available within the "local" RSE (RUCIO Storage Element), i.e., the distributed storage that should exists on each of the data centers. (And that should be mounted when you are logged in).

Option 1 takes much more time that option 2. Furthermore, you would need to keep an internet connection open during the whole download.

Therefore, we should go with option 2.

I can already create a small bash script for VEGA that simlinks all the dataset files into a txt file, that we would need to adapt for each of the data centers. Step by step ;-).

Let me know where I can add this script within `itwinai`.

matbun:
I have created a new tutorial folder on a new branch: https://github.com/interTwin-eu/itwinai/tree/156-easily-access-datasets-on-rucio-data-lake/tutorials/data-lake/pull-dataset

@garciagenrique could you please add an example of "option 2" with some documentation? The goal is giving such example to the interTwin use cases, so that they can reproduce it for their datasets. Perhaps a couple of links to Rucio docs would help as well.

Thanks!

matbun:
[...]

PR #503, opened 2026-08-10
RUCIO dockerfile

okrochak:
# Summary

This is a small PR to bring some of the developments from RI-SCALE SUC1 to the main itwinai repository. I want to share the Dockerfile we've developed together with RUCIO developers that allows for RUCIO usage. This can be helpful for the future use case developers. 
---

**Related issue :** #156


docs/index.rst
```diff
@@ -210,7 +212,7 @@ If you use **itwinai** in your research, please cite:
 
    Bunino et al., (2026). itwinai: A Python Toolkit for Scalable Scientific
    Machine Learning on HPC Systems. Journal of Open Source Software, 11(117),
-   9409. https://doi.org/10.21105/joss.09409
+   1.    https://doi.org/10.21105/joss.09409
```
matbun: why has this been changed?

docs/getting-started/rucio.rst
```diff
@@ -0,0 +1,54 @@
...
+========================
+
+itwinai provides pre-built container images to facilitate the deployment and scaling of machine
+learning applications. If you need to use `RUCIO <https://rucio.cern.ch/>`_ for your use case,
+you can build your own container from ``env-files/torch/rucio.Dockerfile``, which adds the RUCIO
+clients on top of the ``jlab-slim`` itwinai image. Alternatively, you can use an already
+published image:
+
+- **RUCIO + itwinai image**:
+  ``ghcr.io/okrochak/hypermeteo-downscaling-plugin``
```
matbun: it would be better to reference a container image under itwinai namespace, e.g., `ghcr.io/intertwin-eu/itwinai:TAG`

env-files/torch/rucio.cfg
matbun, line 1: is this file needed? I mean, it is already embedded in `rucio.rst`. To avoid duplication I would suggest removing it.

I see it is referenced by `rucio.Dockerfile`, but a user would need to edit it anyways. In interTwin this file was mounted into the jupyterlab container at container startup with pre-filled fields. This removed entirely the need for a user to manually edit/create such file.

env-files/torch/rucio.Dockerfile
matbun, line 1: Rucio clients are already available in `ghcr.io/intertwin-eu/itwinai:jlab-slim-latest`

Why another Dockerfile is needed?

See: https://github.com/interTwin-eu/itwinai/blob/main/env-files/torch/jupyter/slim.Dockerfile

matbun:
Also, I am working on this to fix the CI: https://github.com/interTwin-eu/itwinai/issues/504

env-files/torch/rucio.cfg
okrochak: Hi Matteo, could you elaborate on how exactly this was done in interTwin? I don't understand where the file was pulled from then. Also I think the user would need to edit this file either way to provide their ORCID ID.

env-files/torch/rucio.Dockerfile
okrochak: Thanks, I will take a look. FYI I will be on holiday for 2 weeks so I will come back to his PR after.

env-files/torch/rucio.cfg
matbun: @andrea-manzi @dciangot do you remember how the rucio cfg got mounted in the jlab container?

docs/index.rst
okrochak: typo

env-files/torch/rucio.Dockerfile
okrochak: Hi @matbun, so I have looked into it again. Indeed I am using this Dockerfile as the base image, because when testing this image on TUBITAK HPC within the RI-SCALE project, I had difficulties using RUCIO. Mainly, gfal-ls library wasn't linked properly, and I think there was some confusion between Rucio using its own Python from its own conda environment vs itwinai python. Perhaps the best way to approach this would be to retest if these difficulties also manifest on another site?

Have you used Rucio with itwinai previously? In particular, downloading datasets within a certain scope?

okrochak: If these issues appear on other sites, probably the best way is to address them directly in the base Dockerfile. If they don't appear I guess we can close this MR, since it's likely a problem on a TUBITAK's side.

matbun: In interTwin we demonstrated itwinai-Rucio integration on Vega. I remember we could mount the Rucio cfg into the container and successfully do `rucio whoami`. I don't remember pulling data, though. I would assume it should work once you are correctly authenticated, but I don't remember if we tested it. I would suggest you reach out to Diego and Andrea.

In any case I agree that you could edit the existing Dockerfile directly

okrochak: Hmm, I could also do whoami quite easily, but for actual data-handling I think the gfal library is needed and I had to change the Dockerfile. I will discuss with them during our next meeting and probably include the updates in the slim.Dockerfile

matbun: Okay, good!

matbun:
@okrochak I saw your last commit but I will wait for you to re-request a review (I don't know if you're done yet). No rush

okrochak:
@matbun @r-sarma I think it's ready for review now. I've implemented changes needed for GFAL library in the torch/slim.Dockerfile and torch/jupyter/slim.Dockerfile and tested them myself. I don't think we need to worry about the rucio.cfg file too much, I've removed it for now but there are instructions on what should be in there in the docs through rucio.rst.

docs/getting-started/rucio.rst
```diff
@@ -0,0 +1,47 @@
...
+       [client]
+       rucio_host = https://ri-scale-server.rucioit.cern.ch
+       auth_host = https://ri-scale-server.rucioit.cern.ch
+       auth_type = oidc
+       account = your-account-name
+       oidc_scope = openid profile eduperson_entitlement offline_access
+       oidc_issuer = https://aai-dev.egi.eu/auth/realms/egi
+
+3. **Follow the OIDC authentication flow**: the first RUCIO command prints a URL that you have
+   to open in your browser to complete the authentication:
```
matbun, line 37: Just curious to know whether this is still the way to go, after you had a chat with Andrea/Diego. IIRC, we were mounting it directly into the container with the right credentials and valid OIDC token

docs/index.rst
```diff
@@ -210,7 +212,7 @@ If you use **itwinai** in your research, please cite:
 
    Bunino et al., (2026). itwinai: A Python Toolkit for Scalable Scientific
    Machine Learning on HPC Systems. Journal of Open Source Software, 11(117),
-   9409. https://doi.org/10.21105/joss.09409
+   9409.    https://doi.org/10.21105/joss.09409
```
matbun, line 215: I would probably remove the added spaces here

matbun (commented):
I left just two small comments, but in general it looks good!

r-sarma (dismissed):
Good PR, I think its good to go.

docs/getting-started/rucio.rst
okrochak: Hi Matteo, Andrea shared this setup with me: https://github.com/interTwin-eu/interlink-flux/blob/main/apps/jhub-dev/values.yaml#L401 (l. 401), but I'm not an expert on kubernetes and not sure how to implement this here directly. The above instructions are basically from Marvin Gajek, my RUCIO contact within the RI-SCALE. Anyways, from most new users I think they would have to contact RUCIO admins to set up an account, so a fully automated way is not possible as far as I understand.

matbun: Yes makes sense, thanks for checking with Andrea!

matbun (approved):
LGTM

PR #505, opened 2026-08-18
Migrate to MLflow 3

### LINKED ISSUE #504: CI: all MLflow tests fail with mlflow 3.14
All MLflow-touching tests fail in CI because `mlflow` 3.14.0 turned the filesystem tracking backend into a hard error. `pyproject.toml` pins `mlflow>=2.17.2` (unpinned upper bound), so CI resolves 3.14.0 and every PR is red regardless of its content.

## Error

```
mlflow.exceptions.MlflowException: The filesystem tracking backend (e.g., './mlruns') is in
maintenance mode and will not receive further updates. Please migrate to a database backend
(e.g., 'sqlite:///mlflow.db') ... If the filesystem backend is required for your workflow,
set `MLFLOW_ALLOW_FILE_STORE=true` to opt out of this exception.
```

Raised from `mlflow/store/tracking/file_store.py:225` whenever a `file://` tracking URI is resolved.

## Failing tests

`7 failed, 92 passed, 10 skipped, 1 error`

```
FAILED tests/loggers/test_loggers.py::test_logger_initialization[mlflow_logger]
FAILED tests/loggers/test_loggers.py::test_logger_initialization[loggers_collection]
FAILED tests/loggers/test_loggers.py::test_mlflow_logger_log
FAILED tests/loggers/test_loggers.py::test_loggers_collection_log
FAILED tests/use-cases/test_mnist.py::test_mnist_train_torch
FAILED tests/use-cases/test_mnist.py::test_mnist_train_torch_lightning
FAILED tests/use-cases/test_3dgan.py::test_3dgan_inference
ERROR  tests/loggers/test_prov4ml.py::test_log_prov_documents
```

The three use-case tests fail as `subprocess.CalledProcessError` from `itwinai exec-pipeline`; the same `MlflowException` is in the child process stderr.

## Reproducer

Any PR run, e.g. #500:
https://github.com/interTwin-eu/itwinai/actions/runs/27823638044/job/82342494469

matbun:
Fixes #504 

`mlflow>=2.17.2` was unbounded, so CI resolved mlflow 3.14+, which rejects the filesystem
tracking backend that itwinai uses by default. Every PR was failing regardless of its content
(e.g. #500, a dependabot action bump). This pins mlflow 3 and ports the code to it.

## Changes

- `pyproject.toml`: `mlflow>=2.17.2` → `mlflow>=3.1,<4`; `uv.lock` regenerated (mlflow 3.15.1).
- `src/itwinai/__init__.py`: default `MLFLOW_ALLOW_FILE_STORE=true`, alongside the existing Ray
  env-var defaults. Set on import, so it also covers Ray workers and the forked GPU-monitor
  processes, and is inherited by the `mlflow ui`/`mlflow server` subprocesses. Users who set the
  variable explicitly still win.
- `src/itwinai/loggers.py`, `kind="model"`: mlflow 3 renamed `log_model`'s `artifact_path` to a
  validated `name` and changed `serialization_format` to default to `pt2`, which raises without
  an `input_example`. Now passes `name=` with `serialization_format="pickle"` (mlflow 2
  behaviour), sanitizes identifiers that mlflow 3 rejects (`/ : . % " '`), and filters `log()`
  kwargs against `log_model`'s signature so extras like `context="training"` (used by
  `use-cases/3dgan/model.py`) no longer reach `torch.save`.
- `tests/loggers/test_mlflow_backend.py`: new regression tests against a real backend rather
  than mocks — the existing tests mock `mlflow.log_metric`, which is why the `log_model` break
  was invisible. Includes a subprocess test with `MLFLOW_ALLOW_FILE_STORE` unset.
- Docs: raw `mlflow ui/server --backend-store-uri` → `itwinai mlflow-ui/mlflow-server --path`,
  which carries the opt-out. Raw invocations would now fail for users.

## Notes

- `joss/reproducibility/` still uses raw `mlflow ui` on purpose. It runs inside the pinned
  published image `itwinai:joss-virgo-experiments` (mlflow 2), so it is unaffected.
- Kept the file store rather than migrating to SQLite: a shared SQLite file on a network
  filesystem cannot use WAL and depends on cluster-coherent `fcntl` locking, which Lustre only
  provides when mounted with `-o flock`. Multi-node users should point `MLFLOW_TRACKING_URI` at
  a tracking server.

### DIFF
--- a/docs/use-cases/radio-astronomy.rst
+++ b/docs/use-cases/radio-astronomy.rst
@@ -35,7 +35,7 @@ Logging with MLflow
 -----------------------------------------------------------------------------------------------
 By default, the `config.yaml` ensures that the MLflow logging is enabled during the training.
 During or after the run, you can launch an MLflow server by executing
-`mlflow server --backend-store-uri mllogs/mlflow` and connecting to `http://127.0.0.1:5000/` 
+`itwinai mlflow-server --path mllogs/mlflow` and connecting to `http://127.0.0.1:5000/` 
 in your browser.
 
 Test suite
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
   "typer>=0.9.0",
   "numpy<2.0.0",
   "wandb>=0.18.7",
-  "mlflow>=2.17.2",
+  "mlflow>=3.1,<4",
   "wheel>=0.45.0",
   "seaborn>=0.13.2",
   "py-cpuinfo>=9.0.0",
@@ -93,7 +93,12 @@ hpo = [
   "hpbandster>=0.7.0",
   "gpy>=1.13.2",
 ]
-modelhub = ["hypha-rpc", "dotenv", "httpx", "hypha-artifact"]
+modelhub = [
+  "hypha-rpc",
+  "dotenv",
+  "httpx",
+  "hypha-artifact; python_version >= '3.11'",
+]
 
 [tool.hatch.metadata]
 allow-direct-references = true
--- a/src/itwinai/__init__.py
+++ b/src/itwinai/__init__.py
@@ -62,6 +62,11 @@ if not plain_logger.hasHandlers():
     plain_logger.addHandler(h)
 plain_logger.propagate = False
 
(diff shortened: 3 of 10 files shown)


r-sarma (approved):
LGTM!

PR #506, opened 2026-08-18
Add Claude Code skill for use cases integration

### LINKED ISSUE #499: Claude skill for use case integration with itwinai
Explore the opportunity of creating a Claude skill describing how to do the integration of a new use case in itwinai, maybe using past use cases examples. The integration should cover:
- distributed training
- HPO
- logging
- scalability report and profiling

This would reduce the adoption barrier for all the users working with Claude Code

matbun:
Closes #499.

Adds a Claude Code skill that walks users through integrating their training code into an itwinai
plugin: scaffolding from the plugin template, porting the trainer, then distributed training,
logging, HPO and the scalability report.

The repo now doubles as a plugin marketplace, so it installs once and works from the user's own
plugin repo rather than only inside an itwinai clone:

```
/plugin marketplace add interTwin-eu/itwinai
/plugin install itwinai
```

What's in it:

- `.claude-plugin/`: marketplace and plugin manifests
- `skills/integrating-a-use-case/`: SKILL.md plus reference files loaded on demand
- docs page under Getting started, plus a tutorial that integrates an FNO surrogate for 2D Darcy
  flow. The starting script is in `tutorials/claude-skill/fno-darcy/`; it generates its own data
  and trains in under a minute on CPU
- `tests/test_claude_plugin.py`: checks the plugin version matches pyproject.toml and that the
  skill's reference links resolve
- CLAUDE.md: which skill files to update when the corresponding itwinai code changes


okrochak:
@matbun next week I should have some time and I would like to review this

matbun:
> @matbun next week I should have some time and I would like to review this

Thanks! That would be nice!

